PR #19, opened 2019-01-11
Checkpointing

BenjaminRueth:
Implemented checkpointing and related tests using the mock pattern.

# Todos

- [x] test subcycling (Do we still get the reference solution, if we use subcycling? I assume no. This would be a nice example case for comparison with waveform relaxation, since with waveform relaxation and linear interpolation, we should get the reference solution.)
- [x] ~~can we solve #20 within this PR?~~ we postpone #20


fenicsadapter/fenicsadapter.py
```diff
@@ -159,46 +164,86 @@ def create_coupling_neumann_boundary_condition(self, test_functions):
...
+        if self._interface.isActionRequired(PySolverInterface.PyActionWriteIterationCheckpoint()):
+            # continue FEniCS computation with u_np1
+            # update checkpoint
+            self._u_cp.assign(u_np1)
+            assert (np.isclose(t + dt, self._t_cp + dt))
+            self._t_cp = t + dt
+            assert (np.isclose(n + 1, self._n_cp + 1))
+            self._n_cp = n + 1
+            # todo we might want to put reading the checkpoint into a function (duplicate code. compare to above)
+            u_n.assign(self._u_cp)  # set u_n to value of (updated)checkpoint
```
uekerman: I think this one should go outside the if-statement. `WriteIterationCheckpoint` is only fired if a new timestep is entered for the first time. But if FEniCS subcycles `ReadIterationCheckpoint` and `WriteIterationCheckpoint` are false. Still, you want that `u_n.assign(u_np1)` and that time and n increases

fenicsadapter/fenicsadapter.py
```diff
@@ -159,46 +164,86 @@ def create_coupling_neumann_boundary_condition(self, test_functions):
...
         self.set_read_field(read_field)
         self.set_write_field(write_field)
         self._precice_tau = self._interface.initialize()
 
         if self._interface.isActionRequired(PySolverInterface.PyActionWriteInitialData()):
-            self._interface.writeBlockScalarData(self._read_data_id, self._n_vertices, self._vertex_ids, self._read_data)
+            self._interface.writeBlockScalarData(self._write_data_id, self._n_vertices, self._vertex_ids, self._write_data)
             self._interface.fulfilledAction(PySolverInterface.PyActionWriteInitialData())
 
         self._interface.initializeData()
```
uekerman: better to put this also into the if-statement above, though not completely wrong (i just also updated the adapter example)

fenicsadapter/fenicsadapter.py
BenjaminRueth: If we want to support subcycling, I think we have to pull lines 216-218 out of this if loop and write something like this:
```
u_n.assign(u_np1)
t = t+dt
n=n+1
```
Be careful: If a checkpoint is read, these lines should **not** be executed! Maybe put them in the `else` of `if self._interface.isActionRequired(PySolverInterface.PyActionReadIterationCheckpoint()):` (line 199)

fenicsadapter/fenicsadapter.py
BenjaminRueth: This means I will just intend https://github.com/precice/fenics-adapter/pull/19#discussion-diff-248446873R234 by one tab. Easy fix :)

uekerman: ah, mistake. don't do that. my fault. leave the `initializaData()` out of the if-case

fenicsadapter/fenicsadapter.py
BenjaminRueth: See https://github.com/precice/fenics-adapter/pull/19/commits/66fd9f98992308a99cfa094b83629478abfe52b8 for a resolution. However, we still have an issue with non-matching subcycling #20

BenjaminRueth: We decided to ignore #20 for this PR

BenjaminRueth:
For testing subcycling, I will implement a test `test_advance_continue`, where

https://github.com/precice/fenics-adapter/blob/66fd9f98992308a99cfa094b83629478abfe52b8/tests/test_fenicsadapter.py#L81

returns `False` in any case. This means: no checkpoint is written and no checkpoint is read. This is exactly the behavior that we need inside of an ongoing subcycling iteration.

PR #22, opened 2019-01-21
Config

RafalKulaga:
Added `config.py` module, which reads configuration data from the `JSON`config file. `fenicsadapter.py` has been adjusted accordingly. Also, docstrings and small modification of `README` informing about the name of the `JSON` config file have been added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Make sure to install
 
 ### Build and install the adapter
 
-Run ``python3 setup.py install`` from your shell.
+Run ``python3 setup.py install`` from your shell. Use ``sudo`` if necessary.
 
 ### Test the adapter
 
@@ -33,6 +33,7 @@ You can also run the tests in `fenics-adapter/tests` from this folder. Example: 
 
 Add ``from fenicsadapter import Coupling`` in your FEniCS code. Please refer to the examples in the [tutorials repository](https://github.com/precice/tutorials) for usage examples.
 
+FEniCS configuration file`s name must be: ``precice-adapter-config.json`` and should be placed in the same directory as the solver in use.
 ## Packaging
 
 To create and install the `fenicsadapter` python package the following instructions were used: https://python-packaging.readthedocs.io/en/latest/index.html.
--- a/fenicsadapter/config.py
+++ b/fenicsadapter/config.py
@@ -2,48 +2,40 @@
 
 import os, sys, json
 
-class Config(object):
+class Config:
     """Handles reading of config. parameters of the fenicsadapter based on JSON
-    configuration file. Initializer calls readJSON() method.
+    configuration file. Initializer calls readJSON() method. Instance attributes
+    can be accessed by provided getter functions.
 
-    :ivar _config_file_name: name of the configuration file
+    :param adapter_config_filename: name of the adapter configuration file
+    :ivar _config_file_name: name of the preCICE configuration file
     :ivar _solver_name : name of the solver
     :ivar _coupling_mesh_name: name of mesh as defined in preCICE config
     :ivar _read_data_name: name of read data as defined in preCICE config
     :ivar _write_data_name: name of write data as defined in preCICE config
-    :func readJSON(): method accessing data in JSON file
     """
 
-    def __init__(self):
-        self.readJSON()
+    def __init__(self, adapter_config_filename):
 
-        self._config_file_name
-        self._solver_name
-        self._coupling_mesh_name
-        self._read_data_name
-        self._write_data_name
+        self._config_file_name = None
+        self._solver_name = None
+        self._coupling_mesh_name = None
+        self._read_data_name = None
+        self._write_data_name = None
 
+        self.readJSON(adapter_config_filename)
 
-    def readJSON(self):
-        """ Reads JSON configuration file in "r" mode and saves the data to
-        the respective instance attributes. Config file name must be:
-        precice-adapter-config.json
+    def readJSON(self, adapter_config_filename):
+        """ Reads JSON adapter configuration file and saves the data to
+        the respective instance attributes.
 
         :var path: stores path to the JSON config file
         :var data: data decoded from JSON files
         :var read_file: stores file path
-        :raise IOError: if open fails
         """
 
-        path =  os.getcwd() + "/" + os.path.dirname(sys.argv[0]) + "/precice-adapter-config.json"
-        print(path)
-
-        try:
-            read_file = open(path, "r")
-        except IOError:
-            print("An error occured trying to open the file.")
-            raise IOError
-
+        path = os.path.join(os.getcwd(), os.path.dirname(sys.argv[0]), adapter_config_filename)
+        read_file = open(path, "r")
         data = json.load(read_file)
 
         self._config_file_name = data["config_file_name"]
@@ -53,3 +45,18 @@ class Config(object):
         self._read_data_name = data["interface"]["read_data_name"]
 
         read_file.close()
+
+    def get_config_file_name(self):
+        return self._config_file_name
+
+    def get_solver_name(self):
+        return self._solver_name
+
+    def get_coupling_mesh_name(self):
+        return self._coupling_mesh_name
+
+    def get_read_data_name(self):
+        return self._read_data_name
+
+    def get_write_data_name(self):
+        return self._write_data_name
--- a/fenicsadapter/config.py
+++ b/fenicsadapter/config.py
@@ -3,7 +3,7 @@
 import os, sys, json
 
 class Config(object):
-    """This class handles the configuration of the fenicsadapter based on JSON
+    """Handles reading of config. parameters of the fenicsadapter based on JSON
     configuration file. Initializer calls readJSON() method.
 
     :ivar _config_file_name: name of the configuration file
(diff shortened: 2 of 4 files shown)


fenicsadapter/fenicsadapter.py
```diff
@@ -55,54 +64,98 @@ def eval(self, value, x):
...
-        ## coupling mesh related quantities will be defined later
-        self._coupling_mesh_vertices = None  # mesh vertices in a format that can be understood by preCICE
-        self._mesh_name = None  # name of mesh as defined in preCICE config
-        self._mesh_id = None  # ID of the coupling mesh created from mesh name
-        self._vertex_ids = None  # ID of vertices, will be filled by preCICE
-        self._n_vertices = None  # number of vertices
+        ## coupling mesh related quantities
+        self._coupling_mesh_vertices = None # initialized later
+        self._mesh_name = self._config._coupling_mesh_name
+        self._mesh_id = None # initialized later
```
BenjaminRueth: I think it is also possible set `_mesh_id` at this point. Currently, it is set in `set_coupling_mesh`.

fenicsadapter/fenicsadapter.py
```diff
@@ -55,54 +64,98 @@ def eval(self, value, x):
...
+        self._vertex_ids = None # initialized later
+        self._n_vertices = None # initialized later
 
-        ## write data related quantities will be defined later (write data is written by this solver to preCICE)
-        self._write_data_name = None  # name of write data as defined in preCICE config
-        self._write_data_id = None  # ID of the data on the coupling mesh created from data name
-        self._write_data = None  # actual data
+        ## write data related quantities (write data is written by this solver to preCICE)
+        self._write_data_name = self._config._write_data_name
+        self._write_data_id = None # initialized later
```
BenjaminRueth: Similar to `_mesh_id`

fenicsadapter/fenicsadapter.py
```diff
@@ -55,54 +64,98 @@ def eval(self, value, x):
...
+        self._write_data = None
 
-        ## read data related quantities will be defined later (read data is read by this solver from preCICE)
-        self._read_data_name = None  # name of read data as defined in preCICE config
-        self._read_data_id = None  # ID of the data on the coupling mesh created from data name
-        self._read_data = None  # actual data
+        ## read data related quantities (read data is read by this solver from preCICE)
+        self._read_data_name = self._config._read_data_name
+        self._read_data_id = None # initialized later
```
BenjaminRueth: Similar to `_mesh_id`

README.md
```diff
@@ -33,6 +33,7 @@ You can also run the tests in `fenics-adapter/tests` from this folder. Example:
 
 Add ``from fenicsadapter import Coupling`` in your FEniCS code. Please refer to the examples in the [tutorials repository](https://github.com/precice/tutorials) for usage examples.
 
+FEniCS configuration file`s name must be: ``precice-adapter-config.json`` and should be placed in the same directory as the solver in use.
```
BenjaminRueth: Better: set this name as default and provide option to use a custom name provided by the user.

fenicsadapter/config.py
```diff
@@ -0,0 +1,55 @@
...
+    :ivar _coupling_mesh_name: name of mesh as defined in preCICE config
+    :ivar _read_data_name: name of read data as defined in preCICE config
+    :ivar _write_data_name: name of write data as defined in preCICE config
+    :func readJSON(): method accessing data in JSON file
+    """
+
+    def __init__(self):
+        self.readJSON()
+
+        self._config_file_name
```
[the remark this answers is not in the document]
RafalKulaga: There is `self.readJSON()` call before, which initializes all instance attributes (this is missing in your example). However, I can initialize them with `None` to avoid confusion, or get rid of them in `__init__`  and let the `readJSON()` create them (but it might be confusing as well).

fenicsadapter/fenicsadapter.py
```diff
@@ -55,54 +64,98 @@ def eval(self, value, x):
...
+    :ivar _n_vertices: number of vertices
+    :ivar _write_data_name: name of write data as defined in preCICE config (from JSON config file)
+    :ivar _write_data_id: ID of the data on the coupling mesh created from data name
+    :ivar _write_data: actual data to write
+    :ivar _read_data_name: name of read data as defined in preCICE config (from JSON config file)
+    :ivar _read_data_id: ID of the data on the coupling mesh created from data name
+    :ivar _read_data: actual data to read
+    :ivar _precice_tau: timestep
+
+    :func convert_fenics_to_precice(): converts FEniCS function, into an array of values
```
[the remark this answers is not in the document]
RafalKulaga: Each function in `fenicsadapter.py` has a doc string. Or do you suggest that listing which methods are in the class in doc string of the class is redundant?

fenicsadapter/fenicsadapter.py
```diff
@@ -55,54 +64,98 @@ def eval(self, value, x):
 
 
 class Adapter(object):
```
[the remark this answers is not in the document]
RafalKulaga: Did not know that, thanks. @BenjaminRueth I will change also your code.

fenicsadapter/fenicsadapter.py
RafalKulaga: Agree.

fenicsadapter/fenicsadapter.py
RafalKulaga: Agree.

fenicsadapter/fenicsadapter.py
RafalKulaga: Agree.

fenicsadapter/config.py
```diff
@@ -0,0 +1,55 @@
+"""This is configuration module of fenicsadapter."""
+
+import os, sys, json
+
+class Config(object):
```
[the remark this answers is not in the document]
RafalKulaga: Agree.

fenicsadapter/config.py
```diff
@@ -0,0 +1,55 @@
...
+        :var read_file: stores file path
+        :raise IOError: if open fails
+        """
+
+        path =  os.getcwd() + "/" + os.path.dirname(sys.argv[0]) + "/precice-adapter-config.json"
+        print(path)
+
+        try:
+            read_file = open(path, "r")
+        except IOError:
```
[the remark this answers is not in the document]
RafalKulaga: Agree.

fenicsadapter/config.py
```diff
@@ -0,0 +1,55 @@
...
+        the respective instance attributes. Config file name must be:
+        precice-adapter-config.json
+
+        :var path: stores path to the JSON config file
+        :var data: data decoded from JSON files
+        :var read_file: stores file path
+        :raise IOError: if open fails
+        """
+
+        path =  os.getcwd() + "/" + os.path.dirname(sys.argv[0]) + "/precice-adapter-config.json"
```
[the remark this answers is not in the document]
RafalKulaga: Agree.

fenicsadapter/fenicsadapter.py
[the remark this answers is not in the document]
RafalKulaga: Agree

README.md
RafalKulaga: Current name and location is consistent with the `openfoam-adapter` approach (which I think doesn`t allow to change the name of `.yml` config files). Also, we would have to add one more parameter regarding the coupling to the solver.

fenicsadapter/config.py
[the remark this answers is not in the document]
BenjaminRueth: As `readJSON` is a quite short function, I would propose to just put all the functionality into `__init__`. Then we avoid confusion with members of the object being declared outside of the `__init__` function.

fenicsadapter/fenicsadapter.py
```diff
@@ -55,54 +64,98 @@ def eval(self, value, x):
...
     def __init__(self):
-        self._solver_name = None  # name of the solver, will be configured later
+        self._config = Config()
 
-        self._interface = None  # coupling interface, will be initialized later
+        self._solver_name = self._config._solver_name
 
-        self._dimensions = None
+        self._interface = PySolverInterface.PySolverInterface(self._solver_name, 0, 1)
+        self._interface.configure(self._config._config_file_name)
```
BenjaminRueth: `self._config._config_file_name` is by [convention](https://stackoverflow.com/a/2064400/5158031) an internal variable of `Config`. There are two possible solutions:

1. make it "public", i.e. remove the leading `_` of `_config_file_name`.
2. write an appropriate getter function. 

I think that the preferrable (though more verbose) option is number 2, since we do not want that `_config_file_name` is changed from the outside.

fenicsadapter/fenicsadapter.py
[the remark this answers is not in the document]
floli: Yes, it's redundant. Which methods are in the class is already known from the code.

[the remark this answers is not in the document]
RafalKulaga: We just had a discussion with Benjamin on how to format the doc string. We will keep them less "formal" and make sure they are as helpful as possible.

fenicsadapter/fenicsadapter.py
RafalKulaga: Agree.

README.md
RafalKulaga: Agree.

PR #29, opened 2019-03-04
using pip for build process

BenjaminRueth:
Moving to `pip` for building. Similar to https://github.com/precice/precice/pull/304

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Make sure to install
 
 ### Build and install the adapter
 
-Run ``python3 setup.py install`` from your shell. Use ``sudo`` if necessary.
+Run ``pip3 setup.py install --user`` from your shell.
 
 ### Test the adapter
 
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Make sure to install
 
 ### Build and install the adapter
 
-Run ``pip3 setup.py install --user`` from your shell.
+Run ``pip3 install --user`` from your shell.
 
 ### Test the adapter
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,4 +25,6 @@ matrix:
       - pip3 install scipy
 
 script: 
+  - mkdir -p precice && echo -e "from setuptools import setup\nsetup(name='precice')" > precice/setup.py
+  - $PY -m pip install precice
   - $PY setup.py test 
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,5 +26,5 @@ matrix:
 
 script: 
   - mkdir -p precice && echo -e "from setuptools import setup\nsetup(name='precice')" > precice/setup.py
-  - $PY -m pip install precice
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -29,7 +29,7 @@ Make sure to install
 
 ### Build and install the adapter
 
-Run ``python3 setup.py install`` from your shell. Use ``sudo`` if necessary.
+Run ``pip3 setup.py install --user`` from your shell.
```
fsimonis: This should be `pip3 install --user .`

BenjaminRueth:
@shkodm do you have an idea, how we can make sure that the unit tests pass in travis? Currently, Travis tells us that the [code is broken](https://travis-ci.org/precice/fenics-adapter/jobs/501682273). This is due to the missing dependency `precice`. 

*Reminder:* The unit tests do not require `precice` due to mocking. However, the dependency `precice` should appear in `setup.py`.

README.md
BenjaminRueth: Thanks!

BenjaminRueth:
I think that this PR can soon be closed:

* building with pip works (`pip3 install --user .`)
* Travis takes care of testing (see [here](https://travis-ci.org/precice/fenics-adapter/builds/526113723?utm_source=github_status&utm_medium=notification))

Open ToDos:

- [ ] The `README.md` has to be updated with respect to the building commands.

BenjaminRueth:
Still open:

- [ ] Travis has problem with the dependency `precice` defined [here](https://github.com/precice/fenics-adapter/blob/c301c718da5cad96d76097c7fc7dc478cdaf3213/setup.py#L16). See [Travis output](https://travis-ci.org/precice/fenics-adapter/jobs/529851059#L464-L465).

@shkodm do you know how to fix this?

fsimonis:
@BenjaminRueth You could create a dummy package for precice:

```yaml
script:
- mkdir -p precice && echo -e "from setuptools import setup\nsetup(name='precice')" > precice/setup.py
- $PY -m pip install precice
- $PY setup.py test
```

BenjaminRueth:
> 
> 
> @BenjaminRueth You could create a dummy package for precice:
> 
> ```yaml
> script:
> - mkdir -p precice && echo -e "from setuptools import setup\nsetup(name='precice')" > precice/setup.py
> - $PY -m pip install precice
> - $PY setup.py test
> ```

Sounds like a good idea. Just copy-pasted your proposed modification in https://github.com/precice/fenics-adapter/pull/29/commits/00bd78276df013fc6f17fabf683e6b05100f721a

BenjaminRueth:
Seems to need some debugging, but this is doable: https://travis-ci.org/precice/fenics-adapter/jobs/530251540#L456-L464

fsimonis:
Looks like pip tries to fetch the package instead of installing it from the precice directory. Can you turn `pip install precice` into `pip install ./precice/` and try again?

PR #45, opened 2019-10-13
Remove error message for force calculation in pure 2D setup

uekerman:
I tested the code also in a pure 2D setup, coupling Nutils to FEniCS.


uekerman:
> If there is a test for a pure 2D case, we can remove this error message.

Well not public yet, but there is a test :)

PR #49, opened 2019-10-15
Fix a few problems for vectorial data

uekerman:
As discussed for documentation purpose. I only commented out things that were wrong and I also deleted some additional things for debugging. Feel free to redo the fix and close this PR.

### DIFF
--- a/fenicsadapter/fenicsadapter.py
+++ b/fenicsadapter/fenicsadapter.py
@@ -92,15 +92,12 @@ class CustomExpression(UserExpression):
             self._dimension -= 1
             coords_y = np.zeros(self._coords_x.shape)
         self._coords_y = coords_y
-        if coords_z is None:
-            self._dimension -= 1
-            coords_z = np.zeros(self._coords_x.shape)
 
         self._coords_y = coords_y
-        self._coords_z = coords_z
         self._vals = vals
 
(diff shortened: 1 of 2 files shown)


BenjaminRueth:
I ran some manual test with `heat.py`:

* `python3 heat.py -d -dT 0.1` works
* `python3 heat.py -d -dT 0.1 -t s` fails, but is also expected to fail (Implicit Euler cannot resolve sinusoidal time dependence)
* `python3 heat.py -d -dT 0.1 -wr 5 3` works
* `python3 heat.py -d -dT 0.1 -wr 5 3 -t q` fails, but also expected to fail (Implicit Euler cannot resolve quadratic time dependence)
* `python3 heat.py -d -dT 0.1 -wr 5 3 -t q -wri quadratic` fails, but also expected to fail (Implicit Euler cannot resolve quadratic time dependence, even if we interpolate quadratically)
* `python3 heat.py -d -dT 0.1 -wr 5 3 -t q -mth tr` fails, but also expected to fail (Trapezoidal Rule cannot resolve quadratic time dependence, if we only interpolate linearly)
* `python3 heat.py -d -dT 0.1 -wr 5 3 -t q -mth tr -wri quadratic` works

PR #54, opened 2019-11-11
fix NP bug with identical edge ids

BenjaminRueth:
Activating NP gives an error, since identical vertex ids are provided to preCICE, if the heat.py example is being run.


fenicsadapter/fenicsadapter.py
```diff
@@ -478,10 +469,18 @@ def set_coupling_mesh(self, mesh, subdomain, use_nearest_projection=True):  # as
...
         self._coupling_mesh_vertices, self._n_vertices = self._extract_coupling_boundary_vertices()
+        print(self._coupling_mesh_vertices)
         self._vertex_ids = np.zeros(self._n_vertices)
         self._interface.set_mesh_vertices(self._mesh_id, self._n_vertices, self._coupling_mesh_vertices.flatten('F'), self._vertex_ids)
+
+        """ Define a mapping between coupling vertices and their IDs in precice"""
+        id_mapping = dict()
+        id_at_x, id_at_y = dict()
+        for i in range(self._n_vertices):
+            id_mapping[(self._coupling_mesh_vertices[i, 0], self._coupling_mesh_vertices[i, 1])] = self._vertex_ids[i]
```
BenjaminRueth: Better use as key vertex object id

fenicsadapter/fenicsadapter.py
```diff
@@ -429,59 +429,61 @@ def _are_connected_by_edge(self, v1, v2):
...
-            vertices_1.append(v1.x(1))
-            vertices_2.append(v2.x(0))
-            vertices_2.append(v2.x(1))
-
-        vertices_1 = np.array(vertices_1)
-        vertices_2 = np.array(vertices_2)
-        vertices1_ids = np.zeros(n)
-        vertices2_ids = np.zeros(n)
+            if v1 is not v2:
+                vertices1_ids.append(id_mapping[(v1.x(0), v1.x(1))])
```
BenjaminRueth: Use `id_mapping[v1]`

fenicsadapter/fenicsadapter.py
```diff
@@ -429,59 +429,61 @@ def _are_connected_by_edge(self, v1, v2):
...
         return vertices1_ids, vertices2_ids
 
-    def set_coupling_mesh(self, mesh, subdomain, use_nearest_projection=False):  # as soon as issue https://github.com/precice/fenics-adapter/issues/53 is fixed change default to use_nearest_projection=True
+    def set_coupling_mesh(self, mesh, subdomain, use_nearest_projection=True):  # as soon as issue https://github.com/precice/fenics-adapter/issues/53 is fixed change default to use_nearest_projection=True
         """Sets the coupling mesh. Called by initalize() function at the
         beginning of the simulation.
         """
         self._coupling_subdomain = subdomain
         self._mesh_fenics = mesh
         self._coupling_mesh_vertices, self._n_vertices = self._extract_coupling_boundary_vertices()
```
BenjaminRueth: Also return vertex objects here, might need to add another function `coupling_mesh_vertices = convert(fenics_vertices)` to convert fenics vertices to preCICE vertex coordinate array.

IshaanDesai:
The tests fail right now because of the commit https://github.com/precice/fenics-adapter/commit/bd10b054605e53444be264b5537b4787bdbe34d6 makes the adapter use nearest-neighbor mapping as default. The [reference output of the system tests](https://github.com/precice/systemtests/tree/master/tests/TestCompose_fe-fe.Ubuntu1804.home/referenceOutput) is generated for non-matching meshes and nearest-projection mapping. So this PR is essentially the real fix for the quick fix done in the commit https://github.com/precice/fenics-adapter/commit/bd10b054605e53444be264b5537b4787bdbe34d6

PR #55, opened 2019-11-27
update version 0.1 -> 0.2

BenjaminRueth:
This PR add support for the newest version of python bindings and drops the old version.

These PRs should be merged together:

- [ ] https://github.com/precice/fenics-adapter/pull/55
- [x] https://github.com/precice/precice/pull/576
- [ ] https://github.com/precice/systemtests/pull/123

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,6 @@ matrix:
       - pip3 install scipy
 
 script: 
-  - mkdir -p precice_future && echo -e "from setuptools import setup\nsetup(name='precice_future')" > precice_future/setup.py
-  - $PY -m pip install ./precice_future/
+  - mkdir -p precice && echo -e "from setuptools import setup\nsetup(name='precice')" > precice/setup.py
+  - $PY -m pip install ./precice/
   - if [ "$TRAVIS_OS_NAME" = "linux" ]; then  $PY setup.py test ; else $PY setup.py test -s tests.test_fenicsadapter ; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,6 @@ matrix:
       - pip3 install scipy
 
 script: 
-  - mkdir -p precice && echo -e "from setuptools import setup\nsetup(name='precice', version="2.0.0")" > precice/setup.py
+  - mkdir -p precice && echo -e "from setuptools import setup\nsetup(name='precice', version='2.0.0')" > precice/setup.py
   - $PY -m pip install ./precice/
   - if [ "$TRAVIS_OS_NAME" = "linux" ]; then  $PY setup.py test ; else $PY setup.py test -s tests.test_fenicsadapter ; fi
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -6,13 +6,13 @@
...
-      version='0.1',
+      version='1.0.0',
       description='preCICE-adapter for the open source computing platform FEniCS ',
       url='https://github.com/precice/fenics-adapter',
       author="Benjamin Rueth",
       author_email='benjamin.rueth@tum.de',
       license='LGPL-3.0',
       packages=['fenicsadapter'],
-      install_requires=['precice_future', 'fenics', 'scipy', 'numpy>=1.13.3'],
+      install_requires=['precice>=2.0.0', 'fenics', 'scipy', 'numpy>=1.13.3'],
```
IshaanDesai: This condition leads to the failure of Linux tests: https://travis-ci.org/precice/fenics-adapter/jobs/617834859?utm_medium=notification&utm_source=github_status Do we wait till preCICE v2.0.0 is released before merging this?

.travis.yml
```diff
@@ -46,6 +46,6 @@ matrix:
       - pip3 install scipy
 
 script: 
-  - mkdir -p precice_future && echo -e "from setuptools import setup\nsetup(name='precice_future')" > precice_future/setup.py
-  - $PY -m pip install ./precice_future/
+  - mkdir -p precice && echo -e "from setuptools import setup\nsetup(name='precice', version="2.0.0")" > precice/setup.py
```
IshaanDesai: preCICE version here is 2.0.0, do we wait till v2 release or merge this right now and bump the version later?

BenjaminRueth: The version of the fenics adapter that we want to merge will only be compatible with preCICE 2.0.0, since otherwise the names and everything of the python bindings do not match. Therefore, I would keep the 2.0.0 here. Otherwise, we suggest a compatibility with older, incompatible versions.

setup.py
BenjaminRueth: Fixed in d42c09e. In `.travis.yml` we create a mocked package `precice` with version `2.0.0`. This is only an empty fake version of preCICE that we provide for the sake of fulfilling the requirements. We do not need the actual implementation, since we mock it away, anyway.

The failure was mainly due to a syntactical error.

BenjaminRueth:
The [current failure](https://travis-ci.org/precice/systemtests/jobs/618100791?utm_medium=notification&utm_source=github_status) of the systemtests is expected. As soon as https://github.com/precice/precice/pull/576 is merged, it should be gone, since this makes the newest version of the bindings available for the test. 

@fsimonis Which order for merging PRs do you propose?

fsimonis:
`develop` in adapter repo should always work with `develop` on the main repo, thus I propose to merge in precice/precice first.

.travis.yml
IshaanDesai: Oh okay yes then I agree that we keep 2.0.0 here!

Issue #33, opened 2019-04-26
Refactor adapter for clearer control flow


BenjaminRueth:
https://github.com/precice/fenics-adapter/blob/1b8a278df28330b3354b443e5813f73438fe6200/fenicsadapter/fenicsadapter.py#L244-L248

--> Create Function `restore_solver_state_from_checkpoint`

---

https://github.com/precice/fenics-adapter/blob/1b8a278df28330b3354b443e5813f73438fe6200/fenicsadapter/fenicsadapter.py#L250-L252

--> Create Function `advance_solver_state`

---

https://github.com/precice/fenics-adapter/blob/1b8a278df28330b3354b443e5813f73438fe6200/fenicsadapter/fenicsadapter.py#L255-L261

--> Create Function `save_solver_state_to_checkpoint`

---

https://github.com/precice/fenics-adapter/blob/1b8a278df28330b3354b443e5813f73438fe6200/fenicsadapter/fenicsadapter.py#L254

--> add assertion after this line `assert(not solver_state_has_been_restored_before)`, since this control flow is invalid.

---

https://github.com/precice/fenics-adapter/blob/1b8a278df28330b3354b443e5813f73438fe6200/fenicsadapter/fenicsadapter.py#L233

--> add `if is_read_data_available():`

---

https://github.com/precice/fenics-adapter/blob/1b8a278df28330b3354b443e5813f73438fe6200/fenicsadapter/fenicsadapter.py#L235

--> add `if is_write_data_required():`

---

https://github.com/precice/fenics-adapter/blob/1b8a278df28330b3354b443e5813f73438fe6200/fenicsadapter/fenicsadapter.py#L282-L289

is code duplication of 

https://github.com/precice/fenics-adapter/blob/1b8a278df28330b3354b443e5813f73438fe6200/fenicsadapter/fenicsadapter.py#L235

https://github.com/precice/fenics-adapter/blob/1b8a278df28330b3354b443e5813f73438fe6200/fenicsadapter/fenicsadapter.py#L254-L261

BenjaminRueth:
Use the names from this flowchart: [usage.pdf](https://github.com/precice/fenics-adapter/files/3121702/usage.pdf)

uekerman:
As discussed between @BenjaminRueth and me: The flowchart still has several inconsistencies compared to the adapter example in the wiki that we need to fix.

BenjaminRueth:
Issue is closed by #59.

PR #59, opened 2019-12-11
Offers more control to the user via new interface

### LINKED ISSUE #37: Expose mid-level functions to user
The fenics-adapter offers a very high level interface. This allows to have minimal code changes in existing examples (see `heat.py`), but also limits the user. The following functions can wrap preCICE functionality, while allowing the user to access preCICE at a low level:

```python
def read_block_scalar_data(self._read_data_id, mesh, V):
    precice.read_block_scalar_data(self._read_data_id, self._n_vertices, self._vertex_ids, self._read_data)
    # read_data -> CustomExpression
    return CustomExpression

def write_block_scalar_data(self._write_data_id, u):
    # u -> write_data
    precice.write_block_scalar_data(self._write_data_id, self._n_vertices, self._vertex_ids, self._write_data)

def set_mesh(mesh, SubDomain: coupling_boundary):
    precice.setMeshEdge(...)
    precice.setMeshVertices(...)

#preprocessing

mesh = UnitSquareMesh(10, 10)
m_f = MeshFunction(size_t, mesh, 1)
subdomain.mark(m_f, marker)

def set_mesh(mesh, int: marker, m_f):
    # mesh, marker, m_f ->  coupling_mesh_vertices, n_vertices
    precice.setMeshEdge(...)
    precice.setMeshVertices(...)
```

### LINKED ISSUE #66: Consistently use numpy style docstring
See https://numpydoc.readthedocs.io/en/latest/format.html

BenjaminRueth:
closes #37 and #33
Also closes #66 as the new design now has numpy docstring style documentation


BenjaminRueth:
A usage example is provided in https://github.com/precice/tutorials/tree/introduceMidLevelInterface/HT/partitioned-heat/fenics-fenics-core

fenicsadapter/fenicsadapter.py
```diff
@@ -299,458 +73,236 @@ def __init__(self, adapter_config_filename='precice-adapter-config.json',
...
-                return -test_functions * dot(n, self._coupling_bc_expression) * dolfin.ds
-            else:
-                raise Exception("invalid!")
-        else: # For multiple Neumann BCs integration should only be performed over the respective domain.
-            # TODO: fix the problem here
-            raise Exception("Boundary markers are not implemented yet")
-            return dot(self._coupling_bc_expression, test_functions) * self.dss(boundary_marker)
-        
-    def create_force_boundary_condition(self, function_space):
+    def create_force_boundary_condition(self, Dirichlet_Boundary, function_space):
```
BenjaminRueth: I think we should also try to keep this interface a bit more abstract. Directly return the point sources to the user that are generated from data. This does not necessarily need to be a force. I can also take care of this, when I'm updating the FSI tutorials (see https://github.com/precice/tutorials/pull/62#pullrequestreview-356823828).

fenicsadapter/fenicsadapter.py
```diff
@@ -299,458 +73,236 @@ def __init__(self, adapter_config_filename='precice-adapter-config.json',
...
+                # z is the dead direction so it is supposed that the data is close to zero
+                np.testing.assert_allclose(precice_read_data[:, 2], np.zeros_like(precice_read_data[:, 2]), )
+                assert (np.sum(np.abs(precice_read_data[:, 2])) < 1e-10)
+            else:
+                raise Exception("Dimensions don't match.")
         else:
-            raise Exception("Cannot handle data type %s" % type(data))
-            
-    def _write_block_data(self):
+            raise Exception("Rank of function space is neither 0 nor 1")
```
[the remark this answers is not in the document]
IshaanDesai: I tried to somehow incorporate this idea here: https://github.com/precice/fenics-adapter/pull/59/commits/1ea4f629d111e9a71dd9018b7509b77be53c739c
Note: I am still running into some FEniCS errors with this commit but would like your inputs on the general idea. I suspect there are still some fundamental open discussions

fenicsadapter/fenicsadapter.py
IshaanDesai: I agree that more abstraction is necessary here, especially such that finally the mid-level interface should be able to solve CHT, HT and FSI cases with no special implementation. Currently I have concentrated on getting https://github.com/precice/tutorials/pull/62 running which is a CHT tutorial. Once this is running I plan to incorporate the FSI tutorial with your help. I hope this workflow is okay.

BenjaminRueth:
I realized that in the dependencies you are still using `precice`, it should be `pyprecice`. See 332b7080c8d0ea9cfa0497a40cd6459d9a5fb151. If you want, just `cherry-pick` the commit.

IshaanDesai:
> I realized that in the dependencies you are still using `precice`, it should be `pyprecice`. See [332b708](https://github.com/precice/fenics-adapter/commit/332b7080c8d0ea9cfa0497a40cd6459d9a5fb151). If you want, just `cherry-pick` the commit.

Seems like a very minor change, will just include in one of my commits.

IshaanDesai:
Pending tasks:
- [ ] Fix mock tests
- [ ] Numpy style docstring documentation for all adapter functions

These tasks will not affect the main API design directly so I simultaneously welcome reviews from @BenjaminRueth @uekerman on the design.

fenicsadapter/fenicsadapter.py
```diff
@@ -259,501 +38,308 @@ def __init__(self, adapter_config_filename='precice-adapter-config.json',
...
         self._interface.initialize_data()
 
+        read_data = None
         if self._interface.is_read_data_available():
-            self._read_block_data()
+            read_data = self.read()
+
+        return read_data
+
+    def _set_coupling_mesh(self, mesh, subdomain):
```
BenjaminRueth, line 245: Do we need this function? Currently, it is only called once by `set_coupling_mesh`. Inlining this functions might decrease overall complexity of the class.

fenicsadapter/fenicsadapter.py
```diff
@@ -259,501 +38,308 @@ def __init__(self, adapter_config_filename='precice-adapter-config.json',
         self._interface = precice.Interface(self._solver_name, self._config.get_config_file_name(), 0, 1)
         self._dimensions = self._interface.get_dimensions()
```
BenjaminRueth, line 39: We might consider removing this member and always call `self._interface.get_dimensions()`.

fenicsadapter/fenicsadapter.py
```diff
@@ -259,501 +38,308 @@ def __init__(self, adapter_config_filename='precice-adapter-config.json',
...
 
-        self._coupling_subdomain = None  # initialized later
-        self._mesh_fenics = None  # initialized later
-        self._coupling_bc_expression = None  # initialized later
+        # FEniCS related quantities
         self._fenics_dimensions = None  # initialized later
```
BenjaminRueth, line 42: Add comment: Initialized in `initialize()`. Even nicer: Entirely get rid of this member and determine the dimension (from input data?) when needed. But if this is too troublesome, leave it as it is.

BenjaminRueth: See also my comment above on merging object creation via `__init__`, `set_interpolation_type(...)` and `initialize(...)` above.

fenicsadapter/fenicsadapter.py
```diff
@@ -1,256 +1,35 @@
...
-
-class Adapter:
-    """Initializes the Adapter. Initalizer creates object of class Config (from
-    config.py module).
-
-    :ivar _config: object of class Config, which stores data from the JSON config file
-    """
-    def __init__(self, adapter_config_filename='precice-adapter-config.json',
-                 interpolation_strategy=GeneralInterpolationExpression):
+    def __init__(self, adapter_config_filename='precice-adapter-config.json'):
```
BenjaminRueth, line 32: I am not so happy with the way how the constructor works: The Object has many members that should be declared in `__init__`. However many of these members have to be initialized as `None`, since we cannot give them a reasonable value yet.

My proposal: Merge the functions `initialize(...)` and `set_interpolation_type(...)` into `__init__`. This will allow us to directly initialize the members of the object. If there is no use-case where thing have to be done after `__init__` but before `initialize(...)` is called. We can simply merge the functions. Similar to https://github.com/precice/precice/pull/614.

However, @uekerman might want to comment on this, since it will eliminate the `initialize(...)` function from the adapter, even though it is an important part of the preCICE API that we might want to keep for consistency reasons.

BenjaminRueth: Maybe we could even add (a part of) `initialize_data(...)`? Is it possible to use the adapter *without* calling `initialize_data`? preCICE allows this. Therefore, our adapter should also allow this.

I am not sure about the current restrictions of the adapter. Can we simply not call `initialize_data` and automatically initialize with zeros? I assume some additional information (function space...) might be missing.

tests/MockedPrecice.py
```diff
@@ -32,7 +39,7 @@ def is_action_required(self, py_action):
     def is_coupling_ongoing(self):
         raise Exception("not implemented")
 
-    def configure(self, foo):
+    def mark_action_fulfilled(self, py_action):
```
BenjaminRueth, line 42: `py_action` is misleading. It is not an action of the python action interface that preCICE offers, but simply an action like `action_write_initial_data` etc.

You could simply call it `action`.

fenicsadapter/fenicsadapter.py
IshaanDesai: The variable `self._dimensions` is used multiple times in each time step. Is it advisable to call a preCICE API function multiple times within on time step? The current thinking is that the API function needs to be called only once and the adapter then uses the locally stored value till the end of the coupling.

fenicsadapter/fenicsadapter.py
IshaanDesai: * All the members which are initialized as `None` are used more than once in the Adapter (mostly they are used in every time step). The general strategy followed is that if a member is used only once in an adapter function then it is defined locally within the scope of that function. **The essential trade-off here is to have less adapter member functions and increase preCICE API calls.**
* Currently the user needs to call `initialize_data()` without which the adapter cannot work. I will merge `initialize_data()` into `initialize()` in a way that it is not mandatory to call `initialize_data()` in the workflow

IshaanDesai:
> `fenicsadapter/adapter_core.py` seems a bit chaotic. I am not sure how to reasonably split it up, but we should split it up into a few modules. One is, for sure, the whole `CustomExpression` module. Another one might be everything connected to conversion of formats. The rest might just go into a `misc.py` (not ideal, but maybe we find a better name later)
> 

One more strategy to split `adapter_core.py` is on the lines of applications. With the current design we can state the following distinction:
* FEniCS `Expressions` are used to handle Dirichlet and Neumann coupling boundary conditions for heat transfer based examples.
* FEniCS `PointSources` are used to handle FSI coupling boundary conditions.

fenicsadapter/fenicsadapter.py
BenjaminRueth: This is clearly a conflict runtime vs. readability. Less members decrease complexity, but calling the API too often, might indeed be harmful. Can you do a quick check to find out how time it will cost if we do not store the dimensions in the adapter, but always go through the preCICE API, if we need the dimensions?

I would suggest: If no real runtime problem occurs favour readability. 

Remember: FEniCS' solve is probably much more expensive and we do not really have to worry about this additional API call, at all.

IshaanDesai: I did a time study with respect to problem size with reference to all preCICE API calls which return single values. These calls are: `get_dimensions`, `get_mesh_id`, `get_write_data_id`, `get_read_data_id`. I compared the Global Time when these calls are made each time the respective value is required vs. the Adapter calls these functions once and stores them in local variables which are reusable.

timestudy

No significant time difference is seen for both scenarios and both scenarios scale in almost the same way for increasing problem size.

BenjaminRueth: Thanks for the study! Then I would suggest to get rid of `self._dimensions` and always go over the API function `interface.get_dimensions()`

fenicsadapter/fenicsadapter.py
IshaanDesai: With reference to [this study above](https://github.com/precice/fenics-adapter/pull/59#discussion_r422908918) API related member variables of `__init__ `will be replaced by API calls directly.

IshaanDesai:
* The API function `initializa_data` has been restructured and is now not mandatory while setting up a FEniCS case.
* The adapter function `set_interpolation_strategy` has been removed and the Interpolation Type is now set directly in the adapter constructor.

Moving `initialize` into the adapter constructor will not allow the user to get the `dt` value recommended by preCICE at the time of initialization. The usual API call is:
```
precice_dt = precice.initialize( ... )
```
A constructor cannot return a value and hence another workaround would need to be found to return the correct time step value to the user.

fenicsadapter/fenicsadapter.py
```diff
@@ -1,759 +1,331 @@
...
+        # coupling mesh related quantities
+        self._coupling_mesh_vertices = None  # initialized later
+        self._vertex_ids = None  # initialized later
+        self._n_vertices = None  # initialized later
 
-    def update_boundary_data(self, vals, coords_x, coords_y=None, coords_z=None):
-        """ update the data stored by expression.
+        # read data related quantities (read data is read by use to FEniCS from preCICE)
+        self._read_function_type = None  # stores whether read function is scalar or vector valued
+        self._read_function = None  # Store the FEniCS function (initialized later)
```
BenjaminRueth, line 49: I think we do not need this member anymore. See details further below.

fenicsadapter/fenicsadapter.py
```diff
@@ -1,759 +1,331 @@
...
+        Scalar/Vector write function) read_data is converted.
+        Note: For quasi 2D fenics in a 3D coupled simulation the y component of the vectors is deleted.
+        :return: data read from preCICE in the form of a numpy array with the values like it is used by preCICE
+        """
+        assert (self._read_function_type in list(FunctionType))
 
-    def interpolate(self, x):
-        assert ((self.is_scalar_valued() and self._vals.ndim == 1) or
-                (self.is_vector_valued() and self._vals.ndim == self._dimension))
+        read_data = convert_fenics_to_precice(self._read_function, self._coupling_mesh_vertices)
```
BenjaminRueth, line 129: Here, `self._read_function` is used, but `read_data` is not used a directly being overwritten a few lines below.

fenicsadapter/expression_core.py
```diff
@@ -0,0 +1,213 @@
...
+    def set_boundary_data(self, vals, coords_x, coords_y=None, coords_z=None):
+        """ initialize data stored by expression.
+
+        :param vals: data values on nodes
+        :param coords_x: x coordinates of nodes
+        :param coords_y: y coordinates of nodes
+        :param coords_z: z coordinates of nodes
+        """
+
+        self.update_boundary_data(vals, coords_x, coords_y, coords_z)
```
BenjaminRueth, line 31: I think that this function is unused.

fenicsadapter/adapter_core.py
```diff
@@ -0,0 +1,194 @@
+"""
+This module consists of all core functionality of the FEniCS adapter .
+The module also consists of additional helper functions for the user
```
ajaust, line 3: Why does the module consists of two things? Would it be better to split the module in two parts? Example:
1. One module with the "core functionality"
2. One module with helper functions that can be used by the user and also the adapter implementation?

fenicsadapter/adapter_core.py
```diff
@@ -0,0 +1,194 @@
...
+class FunctionType(Enum):
+    """ Defines scalar- and vector-valued function """
+    SCALAR = 0  # scalar valued function
+    VECTOR = 1  # vector valued function
+
+
+def determine_function_type(input_function):
+    """ Determines if the function is scalar- or vector-valued based on
+    rank evaluation.
+    """
```
ajaust, line 25: `:param` and `:return:` comments are missing?

fenicsadapter/adapter_core.py
```diff
@@ -0,0 +1,194 @@
...
+        x_all, y_all = sample_points[:, 0], sample_points[:, 1]
+        return np.array([data(x, y) for x, y in zip(x_all, y_all)])
+    else:
+        raise Exception("Cannot handle data type %s" % type(data))
+
+
+def get_coupling_boundary_vertices(mesh_fenics, coupling_subdomain, fenics_dimensions, dimensions):
+    """Extracts vertices which lie on the boundary.
+    :return: stack of vertices
+    """
```
ajaust, line 74: Parameters are missing documentation.

fenicsadapter/adapter_core.py
```diff
@@ -0,0 +1,194 @@
...
+    n = 0
+    fenics_vertices = []
+    vertices_x = []
+    vertices_y = []
+    if dimensions == 3:
+        vertices_z = []
+
+    if not issubclass(type(coupling_subdomain), SubDomain):
+        raise Exception("no correct coupling interface defined!")
```
ajaust, line 83: Why is the interface wrong? Could we tell the user in the message what is going wrong more specifically or is it clear from context?

fenicsadapter/adapter_core.py
```diff
@@ -0,0 +1,194 @@
...
+    for v in dolfin.vertices(mesh_fenics):
+        if coupling_subdomain.inside(v.point(), True):
+            n += 1
+            fenics_vertices.append(v)
+            vertices_x.append(v.x(0))
+            if dimensions == 2:
+                vertices_y.append(v.x(1))
+            elif fenics_dimensions == 2 and dimensions == 3:
+                vertices_y.append(v.x(1))
+                vertices_z.append(0)
```
ajaust, line 94: is `fenics_dimensions == 3 and dimensions == 2` a case that could be valid?

fenicsadapter/adapter_core.py
```diff
@@ -0,0 +1,194 @@
...
+    assert (n != 0), "No coupling boundary vertices detected"
+
+    if dimensions == 2:
+        return fenics_vertices, np.stack([vertices_x, vertices_y], axis=1), n
+    elif dimensions == 3:
+        return fenics_vertices, np.stack([vertices_x, vertices_y, vertices_z], axis=1), n
+
+
+def are_connected_by_edge(v1, v2):
+    """Returns true if both vertices are connected by an edge. """
```
ajaust, line 107: Parameters and return value not documented.

fenicsadapter/adapter_core.py
```diff
@@ -0,0 +1,194 @@
...
+            n += 1
+            fenics_vertices.append(v)
+            vertices_x.append(v.x(0))
+            if dimensions == 2:
+                vertices_y.append(v.x(1))
+            elif fenics_dimensions == 2 and dimensions == 3:
+                vertices_y.append(v.x(1))
+                vertices_z.append(0)
+            else:
+                raise Exception("Dimensions do not match!")
```
ajaust, line 96: Dimensions of what do not match? Could we tell the user in the message what is going wrong more specifically or is it clear from context?

fenicsadapter/adapter_core.py
```diff
@@ -0,0 +1,194 @@
...
+    return False
+
+
+def get_coupling_boundary_edges(mesh_fenics, coupling_subdomain, id_mapping):
+    """Extracts edges of mesh which lie on the boundary.
+    :return: two arrays of vertex IDs. Array 1 consists of first points of all edges
+    and Array 2 consists of second points of all edges
+
+    NOTE: Edge calculation is only relevant in 2D cases.
+    """
```
ajaust, line 121: Function parameters are not documented.

Why is it only relevant in 2D? Should this be reflected by the name (`get_coupling_boundary_edges_2D`)? What happens if it is called in 3D?

fenicsadapter/adapter_core.py
```diff
@@ -0,0 +1,194 @@
...
+    return vertices1_ids, vertices2_ids
+
+
+def get_forces_as_point_sources(fixed_boundary, function_space, coupling_mesh_vertices, data):
+    """
+    Creates 2 dicts of PointSources that can be applied to the assembled system.
+    Applies filter_point_source to avoid forces being applied to already existing Dirichlet BC, since this would
+    lead to an overdetermined system that cannot be solved.
+    :return: Returns lists of PointSources
+    """
```
ajaust, line 155: Function parameters are not documented.

fenicsadapter/adapter_core.py
```diff
@@ -0,0 +1,194 @@
...
+    # PointSources are scalar valued, therefore we need an individual scalar valued PointSource for each dimension in a vector-valued setting
+    # TODO: a vector valued PointSource would be more straightforward, but does not exist (as far as I know)
```
ajaust, line 157: Move the todo somewhere else so it does not get lost/forgotten?

fenicsadapter/adapter_core.py
```diff
@@ -0,0 +1,194 @@
...
+
+    return x_forces.values(), y_forces.values()  # don't return dictionary, but list of PointSources
+
+
+def get_coupling_boundary_coordinates(coupling_vertices, fenics_dimensions, dimensions):
+    """Extracts the coordinates of vertices that lay on the boundary. 3D
+    case currently handled as 2D.
+
+    :return: x and y cooridinates.
+    """
```
ajaust, line 185: Function parameters are not documented.
Typo:  cooridinates -> coordinates

fenicsadapter/expression_core.py
```diff
@@ -0,0 +1,213 @@
...
+
+        self._f = self.create_interpolant()
+
+        if self.is_scalar_valued():
+            assert (self._vals.shape == self._coords_x.shape)
+        elif self.is_vector_valued():
+            assert (self._vals.shape[0] == self._coords_x.shape[0])
+
+    def interpolate(self, x):
+        # TODO: the correct way to deal with this would be using an abstract class. Since this is technically more complex and the current implementation is a workaround anyway, we do not use the proper solution, but this hack.
```
ajaust, line 64: Move the todo somewhere else so it does not get lost/forgotten?

fenicsadapter/expression_core.py
```diff
@@ -0,0 +1,213 @@
...
+        """ Interpolates at x. Uses buffered interpolant self._f.
+
+        :return: returns a list containing the interpolated values. If scalar function is interpolated this list has a
+        single element. If a vector function is interpolated the list has self._dimensions elements.
+        """
```
ajaust, line 69: Function parameters are not documented.

fenicsadapter/expression_core.py
```diff
@@ -0,0 +1,213 @@
...
+
+    def eval(self, value, x):
+        """ Evaluates expression at x using self.interpolate(x) and stores result to value.
+
+        :param x: coordinate where expression has to be evaluated
+        :param value: buffer where result has to be returned to
+        """
+        return_value = self.interpolate(x)
+        for i in range(self._vals.ndim):
+            value[i] = return_value[i]
```
ajaust, line 91: Would it not be more pythonic to return the value array?

fenicsadapter/expression_core.py
```diff
@@ -0,0 +1,213 @@
...
+        """ Interpolates at x. Uses buffered interpolant self._f.
+
+        :return: returns a list containing the interpolated values. If scalar function is interpolated this list has a
+        single element. If a vector function is interpolated the list has self._dimensions elements.
+        """
+        raise Exception("Please use one of the classes derived from this class, that implements an actual strategy for"
+                        "interpolation.")
+
+    def create_interpolant(self, x):
+        # TODO: the correct way to deal with this would be using an abstract class. Since this is technically more complex and the current implementation is a workaround anyway, we do not use the proper solution, but this hack.
```
ajaust, line 74: Move the todo somewhere else so it does not get lost/forgotten?

fenicsadapter/expression_core.py
```diff
@@ -0,0 +1,213 @@
...
+            raise Exception("Dimension of the function is 0 or negative!")
+
+
+class GeneralInterpolationExpression(CustomExpression):
+    """Uses RBF interpolation for implementation of CustomExpression.interpolate. Allows for arbitrary coupling
+    interfaces, but has limited accuracy.
+    """
+
+    def create_interpolant(self):
+        interpolant = []
```
ajaust, line 124: Document function including the return value?

fenicsadapter/expression_core.py
```diff
@@ -0,0 +1,213 @@
...
+                interpolant.append(Rbf(self._coords_x, self._coords_y, self._coords_z, self._vals[:, 2].flatten()))
+            else:
+                raise Exception("Problem dimension and data dimension not matching.")
+        else:
+            raise Exception("Dimension of the function invalid/not supported.")
+
+        return interpolant
+
+    def interpolate(self, x):
+        assert ((self.is_scalar_valued() and self._vals.ndim == 1) or
```
ajaust, line 155: Document function including the return value?

fenicsadapter/expression_core.py
```diff
@@ -0,0 +1,213 @@
...
+class ExactInterpolationExpression(CustomExpression):
+    """Uses cubic spline interpolation for implementation of CustomExpression.interpolate. Only allows interpolation on
+    coupling that are parallel to the y axis, and if the coordinates in self._coords_y are ordered such that the nodes
+    on the coupling mesh are traversed w.r.t their connectivity.
+    However, this method allows to exactly recover the solution at the coupling interface, if it is a polynomial of
+    order 3 or lower.
+    See also https://github.com/precice/fenics-adapter/milestone/1
+    """
+
+    def create_interpolant(self):
```
ajaust, line 181: Document function including the return value?

fenicsadapter/expression_core.py
```diff
@@ -0,0 +1,213 @@
...
+                    interp1d(self._coords_y, self._vals[:, 1].flatten(), bounds_error=False, fill_value="extrapolate",
+                             kind="cubic"))
+            else:
+                raise Exception("Problem dimension and data dimension not matching.")
+        else:
+            raise Exception("Dimension of the function is invalid/not supported.")
+
+        return interpolant
+
+    def interpolate(self, x):
```
ajaust, line 201: Document function including the return value?

tests/MockedPrecice.py
```diff
@@ -1,20 +1,24 @@
 from unittest.mock import MagicMock
```
ajaust, line 1: Do the functions here need documentation?

tests/test_fenicsadapter.py
```diff
@@ -8,10 +8,12 @@
 
 fake_dolfin = MagicMock()
 
+
 class MockedArray:
```
ajaust, line 12: Do the functions here need documentation?

IshaanDesai:
Thank you for the reivew @ajaust! This is really helpful!
Regarding the issue with documentation, we are going for a [Numpy docstring style](https://numpydoc.readthedocs.io/en/latest/format.html) documentation. I have finished the documentation the latest commit will hopefully address all documentation related comments in one go.

>1. Why is the interpolation type explicitly set in the code and not in the input file as it would be common in the preCICE configuration? Am I missing up different types of interpolations?

I actually already changed this in the `HT/fenics-fenics` tutorial. Now the interpolation type is not set in the python code by in the JSON config file in this way:
```
{
  "solver_name": "HeatDirichlet",
  "config_file_name": "precice-config.xml",
  "interface": {
      "coupling_mesh_name": "DirichletNodes",
      "write_data_name": "Flux",
      "read_data_name": "Temperature",
      "interpolation_type": "cubic_spline"  **<-- Interpolation Type Setting**
    }
}
```
Do you think this is reasonable?

>     2. `initial_data` is created and only used once a few lines later.
>        Could I also call it the following way?
>        ```
>        # Create a FEniCS Expression to define and control the coupling boundary values
>        coupling_expression = precice.create_coupling_expression()
>        precice.update_coupling_expression(coupling_expression, precice.initialize_data(u_D_function, f_N_function, V))
>        ```

I see no problem in doing this. We are having some final discussions regarding the initialization and it might finally go into the constructor but as of now it can be called this way.

> 3. Are the function names `initialize`, `read`, `write` and `advance` seem to be "vague" compared to the other functions of the adapter. The other function names feel more terse (`update_coupling_expression`, `store_checkpoint` etc.). My suggestions would be:
>        
>        * `initialize`->`initialize_coupling`
>        * `write`,`read` -> `write_data`, `read_data`
>        * `advance` -> `advance_coupling`
>          I am not sure whether this is a good idea or not as it also deviates more from "plain" preCICE.

This is really good, we will look into name changes!

> 4. Is it possible to hide the `precice.finalize()` such that the user does not have to take care of it? Could one hide it in the destructor of the object, for example, and check whether `finalize` it has not been called by the user already?

This could be done but it would deviate a bit from the [Adapter Example](https://github.com/precice/precice/wiki/Adapter-Example) which users are expected to take as a template. Do you have a particular reason in mind on why the user should not have to do this?

ajaust:
Thanks to you guys for all the hard work! @IshaanDesai  @BenjaminRueth 

> I actually already changed this in the `HT/fenics-fenics` tutorial. Now the interpolation type is not set in the python code by in the JSON config file in this way:
> 
> ```
> {
>   "solver_name": "HeatDirichlet",
>   "config_file_name": "precice-config.xml",
>   "interface": {
>       "coupling_mesh_name": "DirichletNodes",
>       "write_data_name": "Flux",
>       "read_data_name": "Temperature",
>       "interpolation_type": "cubic_spline"  **<-- Interpolation Type Setting**
>     }
> }
> ```
> 
> Do you think this is reasonable?

This looks good to me. I think I also got confused by the term **interpolation**. I guess this is not the **mapping** of preCICE (nearest-neighbor, nearest-projection...), but the FEniCS internal interpolation for functions/expressions? In my head mapping and interpolation is the same. 

I would suggest to make it very clear to the user that there is another mapping involved here (maybe another name). Maybe that is clear to most of the FEniCS users, but I first thought the interpolation in the fenics-adapter refers to the preCICE mapping. 

If there is the mapping of preCICE being used and the interpolation of FEniCS, is there any interaction? Do I need both? If so, why? Don't I interpolate/map with preCICE already? 

This might be stupid questions that do not have to be solved in the adapter. It might not be able to be solved in the adapter either, but would need some documentation. 

> > 
> > 2. `initial_data` is created and only used once a few lines later.
> >    Could I also call it the following way?
> >    ```
> >    # Create a FEniCS Expression to define and control the coupling boundary values
> >    coupling_expression = precice.create_coupling_expression()
> >    precice.update_coupling_expression(coupling_expression, precice.initialize_data(u_D_function, f_N_function, V))
> >    ```
> > 
> 
> I see no problem in doing this. We are having some final discussions regarding the initialization and it might finally go into the constructor but as of now it can be called this way.

I am in general fine with the current solution. It was not clear if I would have to call `precice.initialize_data` before `precice.update_coupling_expression` or not and if the temporary variable `initial_data` would be needed.

> 
> This is really good, we will look into name changes!
I am happy that you like it. :-)
> 
> 
> This could be done but it would deviate a bit from the [Adapter Example](https://github.com/precice/precice/wiki/Adapter-Example) which users are expected to take as a template. Do you have a particular reason in mind on why the user should not have to do this?

I guess that is my personal preference and possible also something I slightly dislike over the pure preCICE interface. I have an object that is destroyed at the end of its lifetime so I do not see why I have to help it getting destroyed. But I assume that might also due to some interaction with MPI or so.

BenjaminRueth:
> This looks good to me. I think I also got confused by the term **interpolation**. I guess this is not the **mapping** of preCICE (nearest-neighbor, nearest-projection...), but the FEniCS internal interpolation for functions/expressions? In my head mapping and interpolation is the same.
> 
> I would suggest to make it very clear to the user that there is another mapping involved here (maybe another name). Maybe that is clear to most of the FEniCS users, but I first thought the interpolation in the fenics-adapter refers to the preCICE mapping.
> 
> If there is the mapping of preCICE being used and the interpolation of FEniCS, is there any interaction? Do I need both? If so, why? Don't I interpolate/map with preCICE already?
> 
> This might be stupid questions that do not have to be solved in the adapter. It might not be able to be solved in the adapter either, but would need some documentation.

That's not a stupid question at all! The interpolation that we use in the adapter and the interpolation that we use in preCICE have some potential for confusion, since what we are doing in the adapter is to some degree a hack:

preCICE internally performs interpolation, but we cannot access the interpolant through the API, but only samples of the interpolant on the given mesh (via `read_block_scalar_data`). However, our adapter needs the interpolant to create the expression. Therefore, we create another interpolant from the samples - now inside the adapter. This results in the following:

`data on mesh A -> interpolation (in preCICE) -> obtain samples on mesh B (read_block_scalar_data) -> interpolation (in adapter) -> Expression`

This approach is not ideal, but the only thing the API of preCICE allows. Ideally we could use some kind of "shortcut" and do the following:

`data on mesh A -> interpolation (in preCICE) -> shortcut to the interpolant -> Expression`

uekerman:
> preCICE internally performs interpolation, but we cannot access the interpolant through the API, but only samples of the interpolant on the given mesh (via `read_block_scalar_data`). However, our adapter needs the interpolant to create the expression. Therefore, we create another interpolant from the samples - now inside the adapter. This results in the following:
> 
> `data on mesh A -> interpolation (in preCICE) -> obtain samples on mesh B (read_block_scalar_data) -> interpolation (in adapter) -> Expression`
> 
> This approach is not ideal, but the only thing the API of preCICE allows. Ideally we could use some kind of "shortcut" and do the following:
> 
> `data on mesh A -> interpolation (in preCICE) -> shortcut to the interpolant -> Expression`

That's not the ideal solution to me. How should we create the format that FEniCS needs? The better solution would be to find a way to either project our point values on to the shape functions at the interface. Or to directly use the Gauss points and give those values to FEniCS. Both should be rather simple to do in FEM codes, but we have not yet found how to do this in FEniCS.

IshaanDesai:
@ajaust 
> I guess that is my personal preference and possible also something I slightly dislike over the pure preCICE interface. I have an object that is destroyed at the end of its lifetime so I do not see why I have to help it getting destroyed. But I assume that might also due to some interaction with MPI or so.

I think this is a broad discussion regarding the preCICE API itself. Whether we really need `finalize()` or for that matter even `intialize()` (could potentially be shifted into the constructor) is really a big discussion and thank you for pointing it out here! For now I will stick to the current state of the preCICE API and the Adapter Example in the wiki.

@uekerman 
> That's not the ideal solution to me. How should we create the format that FEniCS needs? The better solution would be to find a way to either project our point values on to the shape functions at the interface. Or to directly use the Gauss points and give those values to FEniCS. Both should be rather simple to do in FEM codes, but we have not yet found how to do this in FEniCS.

* Right now we provide two options for the user to convert the point data into FEniCS Expressions:
  1. Cubic Spline interpolation
  2. RBF interpolation
We recommend that for a simple interface (straight line) the user use Cubic Spline interpolation for higher accuracy and for arbitrary interface (curved / non-standard shape) the user use RBF.
* For 2D cases this is just as good as projecting values on a shape function right? Because I think shape functions would also use lower-order polynomials.
* From the design perspective of the adapter the creation of the Expression and its handling inside the adapter is critical because we tried a design where the Expression is created by the user and lives on the user side but that caused problems with convergence and accuracy.

uekerman:
> > I guess that is my personal preference and possible also something I slightly dislike over the pure preCICE interface. I have an object that is destroyed at the end of its lifetime so I do not see why I have to help it getting destroyed. But I assume that might also due to some interaction with MPI or so.
> 
> I think this is a broad discussion regarding the preCICE API itself. Whether we really need `finalize()` or for that matter even `intialize()` (could potentially be shifted into the constructor) is really a big discussion and thank you for pointing it out here! For now I will stick to the current state of the preCICE API and the Adapter Example in the wiki.

In the preCICE API, `initialize` cannot be moved into the constructor as we need to define the mesh in between.
For `finalize`, there is a point. If we did reconsider how to do the preCICE API today this is indeed a valid discussion. But as we already have it, the current design decision in the adapter is to follow preCICE as closely as possible. This can also be discussed, but my feeling is that it could help users that preCICE and the adapters "speak the same language". Some further remarks (those are no 100% arguments):
* There are critical things happening in `finalize` beyond pure destruction. Synchronization between all participants to avoid early exits (which hurt if you run a coupled simulation in a job script), reduction of performance data to the master rank and printing that to files.
* In non-OO bindings (e.g. Fortran) we would always need a `finalize`.

> * `initialize`->`initialize_coupling`
> * `write`,`read` -> `write_data`, `read_data`
> * `advance` -> `advance_coupling`

upvote on `write_data`, `read_data`, downvote on `initialize_coupling`, `advance_coupling`. I am always a bit reluctant to call anything "coupling" as it is the obvious thing. That is already in the name "preCICE" :smile: . Think of MPI, there it is also not `MPI_Initialize_Messaging` or `MPI_Initialize_Parallelization`.  


> > That's not the ideal solution to me. How should we create the format that FEniCS needs? The better solution would be to find a way to either project our point values on to the shape functions at the interface. Or to directly use the Gauss points and give those values to FEniCS. Both should be rather simple to do in FEM codes, but we have not yet found how to do this in FEniCS.

> * Right now we provide two options for the user to convert the point data into FEniCS Expressions:
>       
> 1. Cubic Spline interpolation
> 2. RBF interpolation
>  We recommend that for a simple interface (straight line) the user use Cubic Spline interpolation for higher accuracy and for arbitrary interface (curved / non-standard shape) the user use RBF.
> 
> * For 2D cases this is just as good as projecting values on a shape function right? Because I think shape functions would also use lower-order polynomials.
> 
> * From the design perspective of the adapter the creation of the Expression and its handling inside the adapter is critical because we tried a design where the Expression is created by the user and lives on the user side but that caused problems with convergence and accuracy.

To avoid any misunderstanding: this is definitely beyond this PR. The current solution is good to go. But the long-term goal should not be to extend the preCICE API here, but to find a better solution in FEniCS, in my opinion. 
Maybe to help the adapter user here: Could we choose the interpolation by default and hide this choice for the non-advanced user?

IshaanDesai:
> upvote on write_data, read_data, downvote on initialize_coupling, advance_coupling. I am always a bit reluctant to call anything "coupling" as it is the obvious thing. That is already in the name "preCICE"

Then I finalise the names `read_data`, `write_data`, `advance`, `initialize` and `finalize`.

> To avoid any misunderstanding: this is definitely beyond this PR. The current solution is good to go. But the long-term goal should not be to extend the preCICE API here, but to find a better solution in FEniCS, in my opinion.
Maybe to help the adapter user here: Could we choose the interpolation by default and hide this choice for the non-advanced user?

I will open another issue addressing this after this PR is merged. Currently each FEniCS related tutorial has a JSON configuration file consisting of the recommended interpolation strategy. The understanding is that a user will refer to one of the tutorials before using this Adapter and would then get an idea of which interpolation type to use. A comprehensive description in the form of a wiki will also follow this PR.

Is this enough for non-advanced users?

fenicsadapter/adapter_core.py
IshaanDesai: Currently the adapter can only handle 2D cases in FEniCS. 2D - 3D coupling where the other participant dimension is 3D is done by neglecting the Z-direction. So currently `fenics_dimensions == 3 and dimension == 2` is not valid.

fenicsadapter/fenicsadapter.py
```diff
@@ -1,759 +1,458 @@
...
+        """
+        Create point sources with reference to fixed boundary in a FSI simulation.
+
+        Parameters
+        ----------
+        fixed_boundary : FEniCS domain
+            FEniCS domain consisting of a fixed boundary condition. For example in FSI cases usually the solid body
+            is fixed at one end.
+        data : PointSource
+            FEniCS PointSource data.
```
BenjaminRueth, line 129: What happens, if `data=None` is provided?

fenicsadapter/fenicsadapter.py
```diff
@@ -1,759 +1,458 @@
...
-        elif self._vals.ndim > 1:
-            return False
-        else:
-            raise Exception("Dimension of the function is 0 or negative!")
+        Update values of point sources using data. This function only works for 2D-pseudo3D coupling.
+
+        Parameters
+        ----------
+        data : PointSource
+            FEniCS PointSource data.
```
BenjaminRueth, line 151: Usage Example, would help. Dummy data is in any case sufficient.

fenicsadapter/fenicsadapter.py
```diff
@@ -1,759 +1,458 @@
...
+        return get_forces_as_point_sources(self._Dirichlet_Boundary, self._function_space, self._coupling_mesh_vertices,
+                                           data)
 
-    def is_vector_valued(self):
-        """ Determines if function being interpolated is vector-valued based on dimension of provided vector self._vals.
+    def read_data(self):
+        """
+        Read data from preCICE. Data is generated in an appropriate form depending on the dimensions of the
+        simulation (2D-3D Coupling, 2D-2D coupling or Scalar/Vector write function).
+        Note: For quasi 2D fenics in a 3D coupled simulation the y component of the vectors is deleted.
```
BenjaminRueth, line 167: What does "appropriate format" mean? Usually this function is used together with `update_point_sources` or `update_coupling_boundary_conditions`, correct? Mentioning that "appropriate" data refers to providing data that is fit for use with these functions might help.

fenicsadapter/fenicsadapter.py
```diff
@@ -1,759 +1,458 @@
...
+    def write_data(self, write_function):
         """
+        Writes data to preCICE. Depending on the dimensions of the simulation (2D-3D Coupling, 2D-2D coupling or
+        Scalar/Vector write function) write_data is first converted into a format needed for preCICE.
 
-        assert (self._write_function_type in list(FunctionType))
+        Parameters
+        ----------
+        write_function : FEniCS function
+            A FEniCS function consisting of the data which this participant will write to preCICE in every time step.
```
BenjaminRueth, line 206: I'm not sure whether "FEniCS function" is the correct term (I know, I use it a lot myself). Is [this here](https://fenicsproject.org/docs/dolfin/1.4.0/python/programmers-reference/functions/function/Function.html) the correct function we are talking about? You could write something like `dolfin.functions.function.Function`, then everything is clear.

fenicsadapter/fenicsadapter.py
```diff
@@ -1,759 +1,458 @@
...
-
-        Note: For quasi 2D fenics in a 3D coupled simulation the y component of the vectors is deleted.
+    def initialize(self, coupling_subdomain, mesh, read_function, function_space, dimensions=2):
+        """
+        Initializes the coupling interface and sets up the mesh in preCICE.
+
+        Parameters
+        ----------
+        coupling_subdomain : FEniCS Domain
+            Part of FEniCS Mesh which is the coupling interface.
```
BenjaminRueth, line 239: I assume that the "coupling interface" here is not the same thing like the "coupling interface" five lines above. We should make sure to be exact about wording. Here, the term interface is already occupied by the interface from computer science.

I would suggest to use `coupling_boundary` instead, since we usually talk about a "coupling mesh" (see https://github.com/precice/precice/wiki/Adapter-Example#mesh-and-data-access). However, in the FEniCS adapter we do not require a discrete mesh, but a (continuous) subdomain.

fenicsadapter/fenicsadapter.py
```diff
@@ -1,759 +1,458 @@
...
+        mesh : FEniCS Mesh
+            FEniCS mesh of the complete region.
```
BenjaminRueth, line 241: Refer to [`dolfin.cpp.mesh.Mesh`](https://fenicsproject.org/docs/dolfin/1.5.0/python/programmers-reference/cpp/mesh/Mesh.html). Note: you could also check the documentation of FEniCS how they solve this kind of naming issue.

fenicsadapter/fenicsadapter.py
```diff
@@ -1,759 +1,458 @@
...
-        """Sets the coupling mesh. Called by initalize() function at the
-        beginning of the simulation.
-        """
-        self._coupling_subdomain = subdomain
-        self._mesh_fenics = mesh
-        self._fenics_vertices, self._coupling_mesh_vertices, self._n_vertices = self._extract_coupling_boundary_vertices()
-        self._vertex_ids = self._interface.set_mesh_vertices(self._mesh_id, self._coupling_mesh_vertices)
-
-        """ Define a mapping between coupling vertices and their IDs in precice"""
+        """ Set up mesh in preCICE """
```
BenjaminRueth, line 276:
```suggestion
        # Set up mesh in preCICE
```
Why not comment like this? `"""` is reserved for [docstrings](https://www.python.org/dev/peps/pep-0008/#documentation-strings). There is a nice explanation of docstrings vs. comments [here](https://stackoverflow.com/a/19076001/5158031). I would say this comment above is actually a comment and not a docstring. Note that for a user of the FEniCS adapter the preCICE API calls are not relevant, since they are supposed to be hidden inside the adapter.

fenicsadapter/fenicsadapter.py
```diff
@@ -1,759 +1,458 @@
...
+        self._vertex_ids = self._interface.set_mesh_vertices(self._interface.get_mesh_id(
+            self._config.get_coupling_mesh_name()), self._coupling_mesh_vertices)
+
+        """ Define a mapping between coupling vertices and their IDs in preCICE """
```
BenjaminRueth, line 280:
```suggestion
        # Define a mapping between coupling vertices and their IDs in preCICE
```
See above

fenicsadapter/fenicsadapter.py
```diff
@@ -1,759 +1,458 @@
...
+        for i in range(len(edge_vertex_ids1)):
+            assert (edge_vertex_ids1[i] != edge_vertex_ids2[i])
+            self._interface.set_mesh_edge(self._interface.get_mesh_id(self._config.get_coupling_mesh_name()),
+                                          edge_vertex_ids1[i], edge_vertex_ids2[i])
 
-        :param write_function_init: function on the write field
-        """
-        self._write_function_type = determine_function_type(write_function_init)
-        self._write_data = self._convert_fenics_to_precice(write_function_init)
+        """ Set read functionality parameters """
```
BenjaminRueth, line 292:
```suggestion
        # Set read functionality parameters
```

fenicsadapter/fenicsadapter.py
```diff
@@ -1,759 +1,458 @@
...
+        self._read_function_type = determine_function_type(read_function)
+        self._function_space = function_space
 
-    def _set_read_field(self, read_function_init):
-        """Sets the read field. Called by initalize() function at the
-        beginning of the simulation.
+        return self._interface.initialize()
 
-        :param read_function_init: function on the read field
+    def initialize_data(self, write_function):
```
BenjaminRueth, line 298: `write_function=None` should also be a valid input. Example use cases: One-way-coupling or if only one participant has `initialize=True`.

fenicsadapter/fenicsadapter.py
```diff
@@ -1,759 +1,458 @@
...
-        :return: dolfin.DirichletBC()
         """
-        self._function_space = function_space
-        self._create_coupling_boundary_condition()
-        return dolfin.DirichletBC(self._function_space, self._coupling_bc_expression, self._coupling_subdomain)
 
-    def create_coupling_neumann_boundary_condition(self, test_functions, function_space=None, boundary_marker=None):
-        """Creates the coupling Neumann boundary conditions using
-        create_coupling_boundary_condition() method.
+        if self._interface.is_action_required(action_write_initial_data()):
```
BenjaminRueth, line 314:
```suggestion
        if self._interface.is_action_required(action_write_initial_data()):
            if write_function == None:
                raise Exception("Wrong usage! Participant requires to write initial data, but no write function is provided to initialize_data().")
```

fenicsadapter/fenicsadapter.py
```diff
@@ -1,759 +1,458 @@
...
-        self._interface.mark_action_fulfilled(precice.action_write_iteration_checkpoint())
+        Advances coupling in preCICE.
 
-    def advance(self, write_function, u_np1, u_n, t, dt, n):
-        """Calls preCICE advance function using precice and manages checkpointing.
-        The solution u_n is updated by this function via call-by-reference. The corresponding values for t and n are returned.
+        Parameters
+        ----------
+        dt : double
+            Time step value used for the last iteration.
```
BenjaminRueth, line 373:
```suggestion
    Length of timestep used by the solver.
```

Stolen from [`SolverInterface::advance(...)`](https://github.com/precice/precice/blob/657f7596b1428d5b26c97eeaa9c26c13b4a32d9e/src/precice/SolverInterface.hpp#L124-L143). Might also be worth considering to use the term `computedTimestepLength` for sake of consistency.

fenicsadapter/fenicsadapter.py
```diff
@@ -1,759 +1,458 @@
...
-        :param u_np1: new value of FEniCS solution u_n+1 at time t_n+1 = t+dt
-        :param u_n: old value of FEniCS solution u_n at time t_n = t; updated via call-by-reference
-        :param t: current time t_n for timestep n
-        :param dt: timestep size dt = t_n+1 - t_n
-        :param n: current timestep
-        :return: return starting time t and timestep n for next FEniCS solver iteration. u_n is updated by advance correspondingly.
+        Returns
+        -------
+        max_dt : double
+            Recommended time step value from preCICE.
```
BenjaminRueth, line 378:
```suggestion
            Maximum length of next timestep to be computed by solver.
```

fenicsadapter/fenicsadapter.py
```diff
@@ -1,759 +1,458 @@
...
-        :param coupling_subdomain: domain where coupling takes place
-        :param mesh: fenics mesh
-        :param read_field: function applied on the read field
-        :param write_field: function applied on the write field
-        :param u_n: initial data for solution
-        :param dimension: problem dimension
-        :param t: starting time
-        :param n: time step n
-        :param coupling_marker: boundary marker, can be used for coupling, multiple Neumann boundary conditions are applied
+    def get_solver_name(self):
```
BenjaminRueth, line 390: In preCICE terminology this is a "participant".

fenicsadapter/fenicsadapter.py
```diff
@@ -1,759 +1,458 @@
...
-            self._set_write_field(write_field)
-        self._precice_tau = self._interface.initialize()
-
-        if write_field:
-            if self._interface.is_action_required(precice.action_write_initial_data()):
-                self._write_block_data()
-                self._interface.mark_action_fulfilled(precice.action_write_initial_data())
+    def is_coupling_ongoing(self):
+        """
+        Tag to check if coupling is still going on.
```
BenjaminRueth, line 401:
```suggestion
        Checks if the coupled simulation is still ongoing.
```

BenjaminRueth: Note: For functions that are mainly wrappers of existing preCICE API functions, you could provide a reference to the preCICE API function (in the python interface?).

fenicsadapter/fenicsadapter.py
```diff
@@ -1,759 +1,458 @@
...
                 raise Exception("Dimensions don't match.")
         else:
             raise Exception("Rank of function space is neither 0 nor 1")
 
-    def _read_block_data(self):
-        """ Reads data from preCICE. Depending on the dimensions of the simulation (2D-3D Coupling, 2D-2D coupling or
-        Scalar/Vector write function) read_data is converted.
-
-        Note: For quasi 2D fenics in a 3D coupled simulation the y component of the vectors is deleted.
+    def initialize(self, coupling_subdomain, mesh, read_function, function_space, dimensions=2):
```
BenjaminRueth, line 232:
```suggestion
    def initialize(self, coupling_subdomain, mesh, function_space, read_function=None, dimensions=2):
```
We should at make the `read_function` an optional argument, since one might use explicit coupling, where the FEniCS participant only writes data. Then the user still has to call `initialize`, but cannot provide a `read_function`.

Further below in `initialize` we should ensure that if no `read_function` is provided that actually the user wants to do explicit coupling and did not just forget it. 

For now, if would just raise an exception:
```
if not read_function:
    raise Exception("Please provide a read_function. Explicit coupling is not yet supported for the FEniCS adapter.")
```
We can implement and test explicit coupling at a later point in time, but we should make sure that our interface to the user already supports it.

BenjaminRueth: Note: I addad an issue on this: https://github.com/precice/fenics-adapter/issues/72

fenicsadapter/fenicsadapter.py
```diff
@@ -1,759 +1,458 @@
...
 
 
-class FunctionType(Enum):
-    """ Defines scalar- and vector-valued function """
-    SCALAR = 0  # scalar valued function
-    VECTOR = 1  # vector valued function
+class Adapter:
+    """
+    Main class for the FEniCS Adapter.
+    """
```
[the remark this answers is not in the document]
IshaanDesai: Can you please check the updated description of the Adapter class and finalize it if found okay?

fenicsadapter/fenicsadapter.py
IshaanDesai: I have added the names of the corresponding API functions in their docstring under the section 
```
Notes
-------
```
and also the link to the `python-bindings` file where the API function can be found.

BenjaminRueth:
I just realized that for checkpointing the `n` is actually not mandatory. We use it in the heat tutorial for bookkeeping only. It is actually not required for the tutorial to work correctly. One can also argue whether `t` should be mandatory for a checkpoint. 

From an API design point of view it would be less restrictive to make these parameters optional to not force the user to provide them, if they are not required for the case the user is solving or do not even exist. However, I think we can also solve this later. If a user really has a problem providing `t` and `n` fake values could be provided and this will not harm as far as I know.

fenicsadapter/fenicsadapter.py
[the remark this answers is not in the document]
BenjaminRueth: Looks very good now :+1:

fenicsadapter/fenicsadapter.py
```diff
@@ -55,64 +65,65 @@ def __init__(self, adapter_config_filename='precice-adapter-config.json'):
...
+        Creates a FEniCS Expression in the form of an object of class GeneralInterpolationExpression or
+        ExactInterpolationExpression. The adapter will hold this object till the coupling is on going.
 
         Parameters
         ----------
-        data : numpy.ndarray
-            The coordinates of the vertices. Coordinates of vertices are stored in a
-            numpy array [N x D] where N = number of vertices and D = dimensions of geometry
+        data : array_like
+            The coupling data. A numpy array [N x D] where N = number of vertices and D = dimensions of geometry.
```
BenjaminRueth, line 89:
```suggestion
            The coupling data. A numpy array [N x D] where N = number of vertices and D = dimensions of the data. I.e. for scalar valued data D = 1 and for vector valued data D = dimension of geometry.
```

fenicsadapter/fenicsadapter.py
```diff
@@ -55,64 +65,65 @@ def __init__(self, adapter_config_filename='precice-adapter-config.json'):
...
         except (TypeError, KeyError):  # works with dolfin 2017.2.0
             coupling_expression = self._my_expression(element=self._function_space.ufl_element(), degree=0)
 
-        if not data:
-            data = np.zeros((self._n_vertices, self._fenics_dimensions))
+        n_vertices, _ = self._coupling_mesh_vertices.shape
+
+        if data is None:
+            # standard initialization with zero valued data
+            data = np.zeros((n_vertices, self._fenics_dimensions))
```
BenjaminRueth, line 110: This would **always** create data for vector valued coupling functions. I think we should additionally check the `self._read_function_type`, if we want to treat this case rigurously.

fenicsadapter/fenicsadapter.py
```diff
@@ -370,35 +391,39 @@ def advance(self, dt):
...
+            Maximum length of timestep to be computed by solver.
         """
         self._first_advance_done = True
         max_dt = self._interface.advance(dt)
         return max_dt
 
     def finalize(self):
         """
-        Finalizes the coupling interface. To be called at the end of the simulation
+        Completes the the coupling interface execution. To be called at the end of the simulation.
```
BenjaminRueth, line 407:
```suggestion
        Completes the the coupling interface execution. To be called at the end of the simulation.
        
        Notes
        -----
        Refer finalize() in https://github.com/precice/python-bindings/blob/develop/precice.pyx
```

fenicsadapter/fenicsadapter.py
```diff
@@ -370,35 +391,39 @@ def advance(self, dt):
...
         ----------
         dt : double
-            Time step value used for the last iteration.
+            Length of timestep used by the solver.
 
         Returns
         -------
         max_dt : double
-            Recommended time step value from preCICE.
+            Maximum length of timestep to be computed by solver.
```
BenjaminRueth, line 399:
```suggestion
            Maximum length of timestep to be computed by solver.
            
            Notes
            -----
            Refer advance() in https://github.com/precice/python-bindings/blob/develop/precice.pyx
```

fenicsadapter/fenicsadapter.py
```diff
@@ -370,35 +391,39 @@ def advance(self, dt):
...
         """
         self._interface.finalize()
 
-    def get_solver_name(self):
+    def get_participant_name(self):
         """
         Returns
         -------
         solver_name : string
             Name of the solver.
```
BenjaminRueth, line 416:
```suggestion
            Name of the solver.
            
            Notes
            -----
            Refer get_participant_name() in https://github.com/precice/python-bindings/blob/develop/precice.pyx
```

fenicsadapter/fenicsadapter.py
BenjaminRueth: I think this approach is good. Note that some functions are still missing the link (see review below).

fenicsadapter/fenicsadapter.py
```diff
@@ -1,759 +1,499 @@
...
+        """
+        self._interface.finalize()
 
-        precice_step_complete = self._interface.is_time_window_complete()
+    def get_participant_name(self):
+        """
+        Returns
+        -------
+        solver_name : string
+            Name of the solver.
```
BenjaminRueth, line 416:
```suggestion
            Name of the participant.
```

fenicsadapter/adapter_core.py
```diff
@@ -0,0 +1,273 @@
...
+            vertices1_ids.append(id_mapping[v1.global_index()])
+            vertices2_ids.append(id_mapping[v2.global_index()])
+
+    vertices1_ids = np.array(vertices1_ids)
+    vertices2_ids = np.array(vertices2_ids)
+
+    return vertices1_ids, vertices2_ids
+
+
+def get_forces_as_point_sources(fixed_boundary, function_space, coupling_mesh_vertices, data):
```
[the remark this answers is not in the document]
IshaanDesai: As of now we do not stop the user from supplying input data in 3D (displacements in x, y, and z directions) but rather simply ignore the z-direction and compute forces on the data in x and y directions. For a 2D-3D pseudo-coupling case this is a valid scenario right? In the case of a 2D-3D we already warn the user multiple times about which strategy is being pursued.

fenicsadapter/fenicsadapter.py
IshaanDesai: Added to current design in the form of one Exception:
```
raise Exception("Please provide a read_function. Explicit coupling is not yet supported for the FEniCS adapter.")
```

PR #61, opened 2019-12-19
Fix "Full tests [Linux]" and "Mock tests [Windows]"

BenjaminRueth:
Something seems to have broken recently:

* last successful build: https://travis-ci.org/precice/fenics-adapter/builds/618263355 uses [Python 3.6.8](https://travis-ci.org/precice/fenics-adapter/jobs/618263356#L177)
* first failing build: https://travis-ci.org/precice/fenics-adapter/builds/618292403 uses [Python 3.6.9](https://travis-ci.org/precice/fenics-adapter/jobs/618292405#L181)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ matrix:
     dist: bionic
     virtualenv:
       system_site_packages: true
-    python: "3.6" 
+    python: "3.7.1" 
     env: PY=python3
     before_install:
     - sudo apt-get install software-properties-common
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,7 @@ matrix:
   - os: linux  
     name: "Full tests [Linux]"
     language: python
-    dist: bionic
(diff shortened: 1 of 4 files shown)


BenjaminRueth:
This seems to be a bug in the python version that we are using: https://github.com/python/typing/issues/582

BenjaminRueth:
Using Ubuntu 19.10 does not solve the issue: Travis still only provides [python 3.6.9](https://travis-ci.org/precice/fenics-adapter/jobs/627376246#L184), even though I would expect python 3.7.5 (this is at least what I get in docker.)

We cannot specify a newer python version via `python: "3.x"` in `.travis.yml`, since we require also `system_site_packages: true` due to FEniCS being installed via `apt-get`.

BenjaminRueth:
Using Conda seems to be the best solution to the problem that we have to use a specific python version plus FEniCS.

PR #70, opened 2020-03-09
Read-Only functionality for coupling FEniCS with aste

### LINKED ISSUE #69: Allow One Way Coupling
Currently the adapter expects 2-way coupling: reading and writing. For aste I want to read only, so there are a few small changes needed.

richahert:
This PR closes #69 but we might not merge it since the whole adapter is about to be changed.

### DIFF
--- a/fenicsadapter/fenicsadapter.py
+++ b/fenicsadapter/fenicsadapter.py
@@ -672,8 +672,8 @@ class Adapter:
         """
         return self._fenics_dimensions == 2 and self._dimensions == 3
 
-    def initialize(self, coupling_subdomain, mesh, read_field, write_field, 
-                   u_n, dimension=2, t=0, n=0, dirichlet_boundary=None ):
+    def initialize(self, coupling_subdomain, mesh, u_n, read_field, write_field=None,
+                   dimension=2, t=0, n=0, dirichlet_boundary=None ):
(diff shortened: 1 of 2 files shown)


fenicsadapter/fenicsadapter.py
```diff
@@ -670,8 +672,8 @@ def _can_apply_2d_3d_coupling(self):
         """
         return self._fenics_dimensions == 2 and self._dimensions == 3
 
-    def initialize(self, coupling_subdomain, mesh, read_field, write_field, 
-                   u_n, dimension=2, t=0, n=0, dirichlet_boundary=None ):
+    def initialize(self, coupling_subdomain, mesh, u_n, read_field, write_field=None,
+                   dimension=2, t=0, n=0, dirichlet_boundary=None ):
```
IshaanDesai: @BenjaminRueth the change in order of input parameters for `initialize` will break the system tests as the tutorials and mock tests are designed according to the earlier order. Quick fix PRs there are necessary before merging this.

BenjaminRueth: I would rather stay with the old order of input arguments to avoid the follow-up fixes you are mentioning. This requires the user to explicitly write `None` (for `write_field`), but since this feature is not widely used and we will change the interface soon in #59, this is not too bad.

Note: Make sure to not forget this requirement in #59.

PR #58, opened 2019-12-11
Support parallel runs

BenjaminRueth:
Related to parallel case in https://github.com/precice/tutorials/pull/59

PR #71, opened 2020-05-03
Support parallel runs

BenjaminRueth:
This PR is based on #59 and is an update of #58. Therefore, #59 has to be merged before this PR.

This PR adds the necessary functionality to support parallel FEniCS runs. An example case is provided in `fenics-adapter/prototyping_parallel`

### Running

```
mpirun -np N python3 heat.py -d
```

and 

```
mpirun -np M python3 heat.py -n
```

### Scenarios

We want to make sure that the following scenarios work (replace `N` and `M` from above):

* `N=1`, `M=1`: serial run
* `N>1`, `M=1`: parallel Dirichlet
* `N=1`, `M>1`: parallel Neumann
* `N>1`, `M>1`: both parallel

Note that depending on the setup some processes might have not coupling boundary vertices. This leads to a `CouplingExpression` without any data. We have to make sure that the adapter also works for these cases!

### Todos

- [x] merge #59
- [x] apply changes of tutorial (in `fenics-adapter/prototyping_parallel`) to [tutorials repository](https://github.com/precice/tutorials/tree/develop/HT/partitioned-heat/fenics-fenics).
    - [x] merge https://github.com/precice/tutorials/pull/62
    - [x] create a PR in https://github.com/precice/tutorials for this purpose (https://github.com/precice/tutorials/pull/120)
- [x] make sure that we can reach machine precision for the HT case (currently error checks are commented out!)
- [ ] add a systemtest


fenicsadapter/fenicsadapter.py
```diff
@@ -1,756 +1,379 @@
...
+
+            if MPI.size(MPI.comm_world) == 1:  # nearest-projection is only supported for non-parallel runs
+                edge_vertex_ids1, edge_vertex_ids2 = extract_coupling_boundary_edges(mesh, subdomain, id_mapping)
+
+                """ Set mesh edges in preCICE to allow nearest-projection mapping"""
+                for i in range(len(edge_vertex_ids1)):
+                    assert (edge_vertex_ids1[i] != edge_vertex_ids2[i])
+                    self._interface.set_mesh_edge(self._mesh_id, edge_vertex_ids1[i], edge_vertex_ids2[i])
+            else:
+                print("Parallel FEniCS runs only support nearest neighbor.")
```
BenjaminRueth, line 287: Nearest projection is not supported for parallel runs. I decided to not investigate this feature in depth. Therefore it might be possible to provide connectivity information also for parallel runs quite easily. I do not really know, since I decided to ignore this feature for the first prototype.

prototyping_parallel/heat.py
```diff
@@ -257,18 +257,16 @@ def determine_gradient(V_g, u, flux):
         print('{rank} of {size}:running stupid loop'.format(rank=MPI.rank(MPI.comm_world), size=MPI.size(MPI.comm_world)))
         for vertex in fenics.vertices(mesh):  # TODO: this loop has no real purpose, however, if we do not execute it, the program hangs later on...
             flux(vertex.x(0), vertex.x(1))
```
IshaanDesai: The code hangs if this loop is removed. The problem of hanged state when passing a FEniCS Function to an adapter function has been narrowed down to the case where a FEniCS Vector Function is passed to the adapter and then the evaluation of the function inside the adapter is attempted. The current commit hangs exactly at [this line](https://github.com/precice/fenics-adapter/pull/71/commits/47c79a9a17aa7c564af64cd4ae017751e59a3393#diff-c99807d6a2ee07096320e7eda2417448R106)

prototyping_parallel/heat.py
```diff
@@ -0,0 +1,315 @@
...
+print('{rank} of {size}:running stupid loop'.format(rank=MPI.rank(MPI.comm_world), size=MPI.size(MPI.comm_world)))
+for vertex in fenics.vertices(mesh):  # TODO: this loop has no real purpose, however, if we do not execute it, the program hangs later on...
+    print('{rank} of {size}: evaluates u_D_function at {vertex}'.format(rank=MPI.rank(MPI.comm_world), size=MPI.size(MPI.comm_world), vertex=(vertex.x(0), vertex.x(1))))
+    u_D_function(vertex.x(0), vertex.x(1))
+print('{rank} of {size}:stupid loop done'.format(rank=MPI.rank(MPI.comm_world), size=MPI.size(MPI.comm_world)))
+print('{rank} of {size}:running stupid loop'.format(rank=MPI.rank(MPI.comm_world), size=MPI.size(MPI.comm_world)))
+for vertex in fenics.vertices(mesh):  # TODO: this loop has no real purpose, however, if we do not execute it, the program hangs later on...
+    print('{rank} of {size}: evaluates f_N_function at {vertex}'.format(rank=MPI.rank(MPI.comm_world), size=MPI.size(MPI.comm_world), vertex=(vertex.x(0), vertex.x(1))))
+    f_N_function(vertex.x(0), vertex.x(1))
+print('{rank} of {size}:stupid loop done'.format(rank=MPI.rank(MPI.comm_world), size=MPI.size(MPI.comm_world)))
```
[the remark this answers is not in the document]
IshaanDesai: The latest design of the adapter does not need a FEniCS Function for the `initialize()` call and hence these loops are now removed.

IshaanDesai:
The parallel function evaluation issue was resolved by a solution in FEniCS Discourse, reference link: https://fenicsproject.discourse.group/t/parallel-fenics-run-hangs-at-external-function-evaluation/3543

fenicsadapter/fenicsadapter.py
```diff
@@ -232,44 +285,53 @@ def write_data(self, write_function):
...
+
+        if not self._empty_rank:
+            write_function_type = determine_function_type(write_function)
+            assert (write_function_type in list(FunctionType))
+            n_vertices, _ = self._owned_coords.shape
+            write_data = convert_fenics_to_precice(write_function, self._coupling_subdomain, self._fenics_dimensions,
+                                                   self._interface.get_dimensions())
+            print("Rank {}: Shape of write data = {}".format(self._rank, write_data.shape))
+            if write_function_type is FunctionType.SCALAR:
+                self._interface.write_block_scalar_data(write_data_id, self._vertex_ids, write_data)
```
BenjaminRueth, line 307:
```suggestion
                assert(write_function.function_space().num_sub_spaces() == 0)
                self._interface.write_block_scalar_data(write_data_id, self._vertex_ids, write_data)
```

fenicsadapter/fenicsadapter.py
```diff
@@ -232,44 +285,53 @@ def write_data(self, write_function):
...
+            elif write_function_type is FunctionType.VECTOR:
+                if self._apply_2d_3d_coupling:
```
BenjaminRueth, line 309:
```suggestion
                assert (write_function.function_space().num_sub_spaces() > 0)
                if self._apply_2d_3d_coupling:
```

fenicsadapter/adapter_core.py
```diff
@@ -3,11 +3,13 @@
 """
 
 import dolfin
-from dolfin import SubDomain, Point, PointSource
-from fenics import FunctionSpace, VectorFunctionSpace, Function
+from dolfin import SubDomain, Point, PointSource, vertices
+from fenics import FunctionSpace, Function
```
BenjaminRueth, line 7:
```suggestion
from fenics import FunctionSpace, VectorFunctionSpace, Function, interpolate
```

fenicsadapter/adapter_core.py
```diff
@@ -81,39 +83,42 @@ def filter_point_sources(point_sources, filter_out):
...
+                                                              coupling_subdomain, fenics_dimensions, dimensions)
+    if type(function) is dolfin.Function:
+        func_vector = function.vector().get_local()
+        print("N dof coordinates = {}".format(len(function.function_space().tabulate_dof_coordinates())))
+        print("len func_vector = {}".format(len(func_vector)))
+        return np.array([func_vector[lid] for lid in local_ids])
+
     else:
-        raise Exception("Cannot handle data type %s" % type(data))
+        raise Exception("Cannot handle data type {}".format(type(function)))
```
BenjaminRueth, line 109: I would suggest to move this `Exception` to the top of the function:

```
if type(function) is not dolfin.Function:
    raise Exception("Cannot handle data type {}".format(type(function)))
```

This saves us some indentation in the remainder of the function.

fenicsadapter/adapter_core.py
```diff
@@ -81,39 +83,42 @@ def filter_point_sources(point_sources, filter_out):
...
     Returns
     -------
     array : array_like
         Array of FEniCS function values at each point on the boundary.
     """
-    if type(data) is dolfin.Function:
-        x_all, y_all = sample_points[:, 0], sample_points[:, 1]
-        return np.array([data(x, y) for x, y in zip(x_all, y_all)])
+    _, _, _, _, local_ids, _ = get_coupling_boundary_vertices(function.function_space().mesh(), function.function_space(),
+                                                              coupling_subdomain, fenics_dimensions, dimensions)
```
BenjaminRueth, line 101: I tried to come up with an approach that helps us to deal with any kind of `function` a user provides. Note that the suggested implementation below still has some errors. I assume that I did not use the different `local_ids` correctly.

# Suggested Implementation

We will have to first transform the `function` (might be anything vector/scalar, low/high order) to a first order function:
```
if function.function_space().num_sub_spaces() > 0:  # function space is a VectorFunctionSpace
    linear_space = VectorFunctionSpace(function.function_space().mesh(), "P", 1)
else:
    linear_space = FunctionSpace(function.function_space().mesh(), "P", 1)

interpolating_function = interpolate(function, linear_space)
```
This makes `interpolating_function.vector().get_local()` easy to understand.

After that, we can `reshape` to get an array that is consistent with the format expected by preCICE:
```
if function.function_space().num_sub_spaces() > 0:  # function space is a VectorFunctionSpace
    assert((len(interpolating_function.function_space().tabulate_dof_coordinates()) / dimensions).is_integer())
    n_vertices = int(len(interpolating_function.function_space().tabulate_dof_coordinates()) / dimensions)
    func_vector = func_vector.reshape([n_vertices, dimensions])
else:  # function space is a FunctionSpace
    n_vertices = len(projected.function_space().tabulate_dof_coordinates())
    func_vector = func_vector.reshape([n_vertices, 1])
```
Finally: Sample `func_vector` where it is necessary:
```
return np.array([func_vector[lid, :] for lid in local_ids])
```

This function should return an array of the correct shape. Compare:
https://github.com/precice/fenics-adapter/blob/666d2dd9d941e4d76ba847cfb0bc6d5e94e3986b/fenicsadapter/adapter_core.py#L103

BenjaminRueth (commented):
I tried to run this version of the FEniCS adapter together with the example provided in `fenics-adapter/prototyping_parallel` in a serial fashion.

Here, I observed some problems with the shape of write data that I tried to fix. A non-working suggested implementation is provided below.

Additionally, I added some assertions that are hopefully helpful.

fenicsadapter/adapter_core.py
BenjaminRueth: c9e1719 now provides a working version of the approach from above. Note that it works for the serial case only. I was not able to test the parallel case due to my FEniCS installation (which does not support parallel runs).

fenicsadapter/fenicsadapter.py
BenjaminRueth: Added in a918437

fenicsadapter/fenicsadapter.py
BenjaminRueth: Added in a918437

fenicsadapter/adapter_core.py
```diff
@@ -97,16 +97,34 @@ def convert_fenics_to_precice(function, coupling_subdomain, fenics_dimensions, d
...
+    if type(function) is not dolfin.Function:
+        raise Exception("Cannot handle data type {}".format(type(function)))
+
+    if function.function_space().num_sub_spaces() > 0:  # function space is a VectorFunctionSpace
+        linear_space = VectorFunctionSpace(function.function_space().mesh(), "P", 1)
+    else:
+        linear_space = FunctionSpace(function.function_space().mesh(), "P", 1)
+
     _, _, _, _, local_ids, _ = get_coupling_boundary_vertices(function.function_space().mesh(), function.function_space(),
                                                               coupling_subdomain, fenics_dimensions, dimensions)
```
BenjaminRueth:
```suggestion
    _, _, _, _, local_ids, _ = get_coupling_boundary_vertices(function.function_space(), coupling_subdomain, fenics_dimensions, dimensions)
```

Here, we can avoid duplicate arguments by only providing the function space. This makes the function easier to use.

BenjaminRueth:
Just realized that we have one more 2D-3D case. Did you make sure that removing 2D-3D coupling does not break https://github.com/precice/tutorials/tree/master/CHT/flow-over-plate/buoyantPimpleFoam-fenics ?

IshaanDesai:
> Just realized that we have one more 2D-3D case. Did you make sure that removing 2D-3D coupling does not break https://github.com/precice/tutorials/tree/master/CHT/flow-over-plate/buoyantPimpleFoam-fenics ?

The 2D-3D coupling removal here is done just to simplify the parallelization workflow. Another PR will be opened shortly to independently remove 2D-3D coupling from the adapter. An accompanying PR to restructure the tutorials accordingly is also being prepared.

IshaanDesai:
@BenjaminRueth can you review this PR even if you are the main author? I can formally request your review as you opened the PR.

IshaanDesai (approved):
The parallel design now works and is ready for finalization. All mock tests pass and major clean-up is complete.

BenjaminRueth:
> @BenjaminRueth can you review this PR even if you are the main author? I cannot formally request your review as you opened the PR.

@IshaanDesai I can do a review. I think I cannot (formally) approve it, since we confused github by shifting ownership of this PR from me to you over the last few months. But as long as we do not get confused it should be fine :smirk:

fenicsprecice/adapter_core.py
```diff
@@ -92,96 +92,183 @@ def filter_point_sources(point_sources, filter_out):
...
-    vertices_z = []
 
     if not issubclass(type(coupling_subdomain), SubDomain):
         raise Exception("No correct coupling interface defined! Given coupling domain is not of type dolfin Subdomain")
 
-    for v in dolfin.vertices(mesh_fenics):
+    # Get mesh from FEniCS function space
+    mesh = function_space.mesh()
+
+    # Get coordinates and global IDs of all vertices of the mesh  which lie on the coupling boundary.
```
BenjaminRueth, line 157: Here the wording *"coordinates ... of all vertices"* looks good.

fenicsprecice/adapter_core.py
```diff
@@ -92,96 +92,183 @@ def filter_point_sources(point_sources, filter_out):
...
+    fenics_gids, fenics_coords = [], []
+    for v in vertices(mesh):
+        if coupling_subdomain.inside(v.point(), True):
+            fenics_gids.append(v.global_index())
+            fenics_coords.append([v.x(0), v.x(1)])
+
+    return np.array(fenics_gids), np.array(fenics_coords)
+
+
+def get_owned_coupling_boundary_vertices(function_space, coupling_subdomain):
```
BenjaminRueth, line 168: From looking at the function name and documentation I do not really see a big difference between this function and `get_fenics_coupling_boundary_vertices`. Is it possible to provide one function serving both purposes? My impression:

* `get_fenics_coupling_boundary_vertices` returns `fenics_gids` and `fenics_coords` on `coupling_subdomain` owned by rank.
* `get_owned_coupling_boundary_vertices` returns `fenics_gids`, **`fenics_lids`**, and `fenics_coords`  on `coupling_subdomain` owned by rank.

This is almost the same. Can we not just use `get_owned_coupling_boundary_vertices` everywhere? Is it too costly to get `fenics_lids`, if they are not needed?

fenicsprecice/adapter_core.py
```diff
@@ -92,96 +92,183 @@ def filter_point_sources(point_sources, filter_out):
...
-def get_coupling_boundary_edges(mesh_fenics, coupling_subdomain, id_mapping):
+            dof_nm1 = None  # If function_space is VectorFunctionSpace then each vertex has multiple DoFs
+            for dof in dofs:
+                if (dof == [v.x(0), v.x(1)]).all() and (dof != dof_nm1).any():
+                    owned_gids.append(v.global_index())
+                    owned_lids.append(v.index())
+                    owned_coords.append([v.x(0), v.x(1)])
+                dof_nm1 = dof
+
+    return np.array(owned_gids), np.array(owned_lids), np.array(owned_coords)
```
BenjaminRueth, line 212: Here "vertex" is local ids, global ids and coordinates.

fenicsprecice/adapter_core.py
```diff
@@ -92,96 +92,183 @@ def filter_point_sources(point_sources, filter_out):
...
+            for dof in dofs:
+                if (dof == [v.x(0), v.x(1)]).all() and (dof != dof_nm1).any():
+                    ownership = True
+                    break
+                dof_nm1 = dof
+
+            if ownership is False:
+                unowned_gids.append(v.global_index())
+
+    return np.array(unowned_gids)
```
BenjaminRueth, line 257: Here "vertex" is just global ids.

fenicsprecice/adapter_core.py
```diff
@@ -92,96 +92,183 @@ def filter_point_sources(point_sources, filter_out):
...
+                    owned_lids.append(v.index())
+                    owned_coords.append([v.x(0), v.x(1)])
+                dof_nm1 = dof
+
+    return np.array(owned_gids), np.array(owned_lids), np.array(owned_coords)
+
+
+def get_unowned_coupling_boundary_vertices(function_space, coupling_subdomain):
+    """
+        Extracts vertices which this rank owns from a given function space which lie on the given coupling domain.
```
BenjaminRueth, line 217:
```suggestion
        Extracts global ids of vertices which this rank owns from a given function space which lie on the given coupling domain.
```

BenjaminRueth (commented):
First of all: Thanks a lot for this redesign! It took quite a while, but I really like the result a lot.

I tried to run all FEniCS tutorials using this adapter `HT` worked nicely (only on 2+1 procs though), `CHT` and both `FSI` cases had issues (see https://github.com/precice/tutorials/pull/120#pullrequestreview-547362241).

For testing I cherry-picked the two commits from https://github.com/precice/tutorials/pull/120 onto https://github.com/precice/tutorials/pull/124 since this is the setup that we want to release with this version of the adapter.

I think #102 has to be merged before this PR. Is this correct?

I also observed a slight inconsistency in wording of `vertex` and `coordinates`. I would suggest the following definition: A `vertex` has 

* a `coordinate` (x,y,z)
* a `value` (data living on the vertex)
* a `global_id`
* a `local_id`

I tried to highlight some places in the code below. Often functions say they return "a vertex", but in the end only return some properties of the vertex depending on the use case. Why not create a (very simple) `vertex` object that holds the properties we need? I think this might safe us some confusion in the future. However, since this is a pure refactoring, we can also postpone this, if it is too much work. Most issues can be fixed easily by being a bit more specific in the documentation.

fenicsprecice/adapter_core.py
```diff
@@ -274,3 +354,186 @@ def get_forces_as_point_sources(fixed_boundary, function_space, coupling_mesh_ve
 
     return x_forces.values(), y_forces.values()  # don't return dictionary, but list of PointSources
 
+
+def determine_shared_vertices(comm, rank, function_space, owned_gids, unowned_gids):
```
BenjaminRueth, line 358: Function name is not ideal. Maybe
```suggestion
def determine_shared_vertices_ownership_map(comm, rank, function_space, owned_gids, unowned_gids):
```
And this `map` is the dictionary mapping `global_ids` of shared `vertices` to the owning `rank`?

fenicsprecice/adapter_core.py
```diff
@@ -274,3 +354,186 @@ def get_forces_as_point_sources(fixed_boundary, function_space, coupling_mesh_ve
...
+                if uo_gid == o_gid:
+                    send_data[o_gid] = neigh
+
+    return send_data, recv_data
+
+
+def communicate_shared_vertices(comm, rank, fenics_gids, owned_coords, fenics_coords, owned_data,
+                                send_pts, recv_pts):
+    """
+
```
BenjaminRueth, line 465:
```suggestion
Triggers asynchrous communication between ranks of this solver to exchange data for shared vertices. Rank owning a shared vertex sends latest data to all ranks it is sharing this vertex with.
```

fenicsprecice/adapter_core.py
```diff
@@ -274,3 +354,186 @@ def get_forces_as_point_sources(fixed_boundary, function_space, coupling_mesh_ve
...
+    Returns
+    -------
+    fenics_data : array_like
+        The updated coupling data. A dictionary containing nodal data with vertex coordinates of all seen vertices as
+        the keys, and the associated data as values.
+
+    """
+    # Create fenics type shared data array for communication
+    fenics_data = dict.fromkeys(tuple(key) for key in fenics_coords)
```
BenjaminRueth, line 498: Why do we create this data structure here? Why not just add the new data to `owned_data`?

fenicsprecice/adapter_core.py
```diff
@@ -274,3 +354,186 @@ def get_forces_as_point_sources(fixed_boundary, function_space, coupling_mesh_ve
...
+
+    # Attach data read from preCICE to appropriate ids in FEniCS style array (which includes duplicates)
+    for coord in fenics_coords:
+        for owned_coord in owned_coords:
+            if (coord == owned_coord).all():
+                fenics_data[tuple(coord)] = owned_data[tuple(owned_coord)]
```
BenjaminRueth, line 504: See comment above: We can just skip this, if we directly work with `owned_data` (maybe let's call it `data` then). Or am I missing something?

fenicsprecice/fenicsprecice.py
```diff
@@ -1,15 +1,19 @@
...
-    get_coupling_boundary_vertices, get_coupling_boundary_edges, get_forces_as_point_sources
-from .expression_core import SegregatedRBFInterpolationExpression
+from .adapter_core import FunctionType, determine_function_type, convert_fenics_to_precice, \
+    get_fenics_coupling_boundary_vertices, get_owned_coupling_boundary_vertices, get_unowned_coupling_boundary_vertices, \
+    get_coupling_boundary_edges, get_forces_as_point_sources, determine_shared_vertices, communicate_shared_vertices,\
+    CouplingMode
+from .expression_core import SegregatedRBFInterpolationExpression, EmptyExpression
 from .solverstate import SolverState
 from fenics import Function, FunctionSpace
+from mpi4py import MPI
```
BenjaminRueth, line 16: Why do we use `from mpi4py import MPI` here and `from fenics import MPI` [above](https://github.com/precice/fenics-adapter/pull/71/files#diff-28d189e544b09e294873b3e88f20eb28e6fba8b856511159660c65782292e524R10)? Can we make this consistent either way or the other?

fenicsprecice/fenicsprecice.py
```diff
@@ -44,19 +48,32 @@ def __init__(self, adapter_config_filename='precice-adapter-config.json'):
...
+        self._dofmap = None  # initialized later using function space provided by user
+        self._fenics_gids = None
+        self._fenics_coords = None
+        self._coupling_subdomain = None
 
         # coupling mesh related quantities
-        self._coupling_mesh_vertices = None  # initialized later
+        self._owned_lids = None
+        self._owned_gids = None  # initialized later
+        self._owned_coords = None  # initialized later
```
BenjaminRueth, line 70:
```suggestion
        self._owned_coupling_vertices_local_ids = None
        self._owned_coupling_vertices_global_ids = None  # initialized later
        self._owned_coupling_vertices_coordinates = None  # initialized later
```

Possible refactoring: Collect this under `self._owned_coupling_vertices`.

fenicsprecice/fenicsprecice.py
```diff
@@ -44,19 +48,32 @@ def __init__(self, adapter_config_filename='precice-adapter-config.json'):
...
+        self._interface = precice.Interface(self._config.get_participant_name(), self._config.get_config_file_name(),
+                                            self._rank, self._size)
 
         # FEniCS related quantities
-        self._fenics_dimensions = None
         self._read_function_space = None  # initialized later
+        self._write_function_space = None  # initialized later
+        self._dofmap = None  # initialized later using function space provided by user
+        self._fenics_gids = None
+        self._fenics_coords = None
```
BenjaminRueth, line 64: Possible refactoring: Collect this under `self._fenics_vertices`

fenicsprecice/fenicsprecice.py
```diff
@@ -44,19 +48,32 @@ def __init__(self, adapter_config_filename='precice-adapter-config.json'):
...
+        self._coupling_subdomain = None
 
         # coupling mesh related quantities
-        self._coupling_mesh_vertices = None  # initialized later
+        self._owned_lids = None
+        self._owned_gids = None  # initialized later
+        self._owned_coords = None  # initialized later
+        self._unowned_gids = None
         self._vertex_ids = None  # initialized later
```
BenjaminRueth, line 72:
```suggestion
        self._precice_vertex_ids = None  # initialized later
```
Has nothing to do with FEniCS, correct? These are the preCICE IDs.

fenicsprecice/fenicsprecice.py
```diff
@@ -183,42 +201,31 @@ def read_data(self):
...
 
-        return {tuple(key): value for key, value in zip(vertices, read_data)}
+            owned_read_data = {tuple(key): value for key, value in zip(self._owned_coords, read_data)}
+            # print("Rank {}: Data before communication: {}".format(self._rank, owned_read_data))
+            updated_data = communicate_shared_vertices(self._comm, self._rank, self._fenics_gids,
+                                                       self._owned_coords, self._fenics_coords, owned_read_data,
+                                                       self._to_send_pts, self._to_recv_pts)
+            # print("Rank {}: Data after communication: {}".format(self._rank, updated_data))
+        else:  # if there are no vertices, we return empty data
+            updated_data = None
```
BenjaminRueth, line 222:
```suggestion
            read_data = {tuple(key): value for key, value in zip(self._owned_coords, read_data)}
            read_data = communicate_shared_vertices(self._comm, self._rank, self._fenics_gids,
                                                       self._owned_coords, self._fenics_coords, read_data,
                                                       self._to_send_pts, self._to_recv_pts)
        else:  # if there are no vertices, we return empty data
            read_data = None
```
See comment above. I think appending to `read_data` instead of redefining it inside `communicate_shared_vertices` is the cleaner solution.

fenicsprecice/fenicsprecice.py
```diff
@@ -183,42 +201,31 @@ def read_data(self):
...
+
+        if not self._empty_rank:
+            for key, val in updated_data.items():
+                assert(val is not None), ("Rank{}: {}: {} is not allowed".format(self._rank, key, val))
+
+        return updated_data
```
BenjaminRueth, line 228:
```suggestion
        if not self._empty_rank:
            for key, val in read_data.items():
                assert(val is not None), ("Rank{}: {}: {} is not allowed".format(self._rank, key, val))

        return read_data
```

setup.py
```diff
@@ -22,6 +22,6 @@
       author_email='benjamin.rueth@tum.de',
       license='LGPL-3.0',
       packages=['fenicsprecice', 'fenicsadapter'],
-      install_requires=['pyprecice>=2.0.0', 'fenics', 'scipy', 'numpy>=1.13.3'],
+      install_requires=['pyprecice>=2.0.0', 'fenics', 'scipy', 'numpy>=1.13.3', 'mpi4py'],
```
BenjaminRueth, line 25: See above. `from fenics import MPI` might be sufficient.

BenjaminRueth (commented):
Hit the `Submit review` button too early.

Regarding the `vertex` refactoring one more comment: Feel free to open an issue and link to this review. We can also go for the less object oriented approach now and put the refactoring on top.

setup.py
IshaanDesai: We need `mpi4py` for the non-blocking communication commands, hence keeping this requirement.

fenicsprecice/adapter_core.py
BenjaminRueth: After short discussion: Let's create a copy of `owned_data` called `fenics_data` and then append the missing keys `fenics_data[key] = None`. Due to security reasons, we don't want to modify `owned_data` in-place.

PR #79, opened 2020-08-03
Provide adapter version to user

IshaanDesai:
This PR properly handles providing a version number to the `fenics-adapter`. 
[Method number 3](https://packaging.python.org/guides/single-sourcing-package-version/#single-sourcing-the-version) is used as it is deemed to be the most appropriate one for this python module.


fenicsadapter/version.py
```diff
@@ -0,0 +1,4 @@
+"""
+Maintain fenicsadapter module version number in this file
+"""
+__version__ = '0.2'
```
BenjaminRueth, line 4: There should be a "newline" at the end of the file.

BenjaminRueth:
@IshaanDesai Is there a way to query the version similar to `python3 -c "import numpy; print(numpy.__version__)"?

IshaanDesai:
> @IshaanDesai Is there a way to query the version similar to `python3 -c "import numpy; print(numpy.__version__)"`?

If `__version__="1.0.0"` is added in `fenicsprecice/__init__py` then such a query form is possible. I had not done this in this PR because there was some issue but I tested it right now and it seems to work. This can be added to the adapter now.

PR #81, opened 2020-08-04
Fix data initialization

### LINKED ISSUE #80: Location of read_data in partitioned-heat correct?
Is the current location of read_data correct?

https://github.com/precice/tutorials/blob/b3f92fc54738fd1d006e9ee444d5127b2668cbd5/HT/partitioned-heat/fenics-fenics/heat.py#L206

Note that we initialize our data before that point

https://github.com/precice/tutorials/blob/b3f92fc54738fd1d006e9ee444d5127b2668cbd5/HT/partitioned-heat/fenics-fenics/heat.py#L124-L136

Therefore, I think that `read_data` should go after `advance` 

https://github.com/precice/tutorials/blob/b3f92fc54738fd1d006e9ee444d5127b2668cbd5/HT/partitioned-heat/fenics-fenics/heat.py#L225

maybe with an additional `is_coupling_ongoing` check.

@IshaanDesai, @uekerman Any opinions on this? Am I missing something?

BenjaminRueth:
closes #80 

This PR provides a draft (or a solution?) for #80. 

# Main changes

* Include `initialize_data` in `initialize`
* Make `write_function` optional argument of `initialize`
* Raise assertion, if no `write_function` is provided, but `self._interface.is_action_required(precice.action_write_initial_data())`
* We need some additional logic to be able to call `coupling_expression.is_scalar_valued()` or `coupling_expression.is_vector_valued()` on an empty `coupling_expression`. Empty means, created, but not updated.

# Restrictions

* we do not check `self._interface.is_read_data_available()` anymore.

# Todo

- [x] delete `prototype` folder before merge
- [x] update https://github.com/precice/tutorials accordingly. See https://github.com/precice/tutorials/pull/101


fenicsadapter/fenicsadapter.py
```diff
@@ -75,46 +75,28 @@ def __init__(self, adapter_config_filename='precice-adapter-config.json'):
...
-            # standard initialization with zero valued data
-            if self._read_function_type is FunctionType.SCALAR:
-                data = np.zeros(n_vertices)
-            elif self._read_function_type is FunctionType.VECTOR:
-                data = np.zeros((n_vertices, self._fenics_dimensions))
-            else:
-                raise Exception("No valid read_function is provided in initialization. Cannot create coupling expression")
-
-        self.update_coupling_expression(coupling_expression, data)
+        coupling_expression._function_type = self._read_function_type
```
BenjaminRueth, line 99: Accessing (or creating) the member `coupling_expression._function_type` is a quick-and-dirty solution. We should better us a function call like `coupling_expression.set_function_type(self._read_function_type)` to respect the privacy of our `CouplingExpression`.

fenicsadapter/fenicsadapter.py
```diff
@@ -133,43 +115,17 @@ def update_coupling_expression(self, coupling_expression, data):
         """
         coupling_expression.update_boundary_data(data, self._coupling_mesh_vertices[:, 0], self._coupling_mesh_vertices[:, 1])
 
-    def create_point_sources(self, fixed_boundary, data=None):
+    def create_point_sources(self, fixed_boundary):
```
BenjaminRueth, line 118: This one-liner function looks a bit unnecessary. We should probably rename it. I don't think that we should add the `fixed_boundary` as an argument to `update_point_sources`, since this will force the user to provide the `fixed_boundary` over and over again even though it is always the same boundary (and it also **has to be** the same boundary).

fenicsadapter/fenicsadapter.py
```diff
@@ -133,43 +115,17 @@ def update_coupling_expression(self, coupling_expression, data):
...
-                data = np.zeros(n_vertices)
-            elif self._read_function_type is FunctionType.VECTOR:
-                data = np.zeros((n_vertices, self._fenics_dimensions))
-            else:
-                raise Exception("No valid read_function is provided in initialization. Cannot create point sources")
-
         self._Dirichlet_Boundary = fixed_boundary
-        return self.update_point_sources(data)
 
     def update_point_sources(self, data):
```
BenjaminRueth, line 130: If we remove (or rename) `create_point_source`, we should also change the name of this function. For the `CouplingExpression` the pair of `create/update_coupling_expression` makes sense. For the `PointSource` it does not. Suggestion: `get_point_source(self, data)`. The `fixed_boundary` has to be set elsewhere (see above).

fenicsadapter/fenicsadapter.py
IshaanDesai: `create_point_sources` can be removed and `fixed_boundary` can be included in `initialize()` as an optional argument.

fenicsadapter/fenicsadapter.py
IshaanDesai: `get_points_sources(self, data)` accurately describes what this function does :+1:

fenicsadapter/fenicsadapter.py
```diff
@@ -327,36 +285,16 @@ def initialize(self, coupling_subdomain, mesh, function_space, dimensions=2):
...
-        if write_function:
-            if self._interface.is_action_required(precice.action_write_initial_data()):
-                self.write_data(write_function)
-                self._interface.mark_action_fulfilled(precice.action_write_initial_data())
+        if self._interface.is_action_required(precice.action_write_initial_data()):
+            assert(write_function)  # if action is required, write function MUST be given.
+            self.write_data(write_function)
+            self._interface.mark_action_fulfilled(precice.action_write_initial_data())
 
         self._interface.initialize_data()
```
IshaanDesai, line 295: Making `initialize_data()` a part of `initialize()` means it will be called even if the user wishes to start with all coupling variables equal to zero. Is this acceptable use of the preCICE API?

fenicsadapter/fenicsadapter.py
```diff
@@ -327,36 +276,16 @@ def initialize(self, coupling_subdomain, mesh, function_space, dimensions=2):
...
-        """
-        self._non_standard_initialization = True
+        precice_dt = self._interface.initialize()
 
-        if write_function:
-            if self._interface.is_action_required(precice.action_write_initial_data()):
-                self.write_data(write_function)
-                self._interface.mark_action_fulfilled(precice.action_write_initial_data())
+        if self._interface.is_action_required(precice.action_write_initial_data()):
+            assert write_function  # if action is required, write function MUST be given.
```
BenjaminRueth, line 282: This should be a proper error.

fenicsadapter/fenicsadapter.py
```diff
@@ -268,9 +212,9 @@ def write_data(self, write_function):
         else:
             raise Exception("write_function provided is neither VECTOR nor SCALAR type")
 
-    def initialize(self, coupling_subdomain, mesh, function_space, dimensions=2):
+    def initialize(self, coupling_subdomain, mesh, function_space, dimensions=2, write_function=None, fixed_boundary=None):
```
BenjaminRueth, line 215: dimension should go into constructor

BenjaminRueth: we should also make sure to raise an Exception, if `dimension != 2` as long as #1 is not resolved.

fenicsadapter/fenicsadapter.py
IshaanDesai: Resolved in discussion.

IshaanDesai:
As we no longer return data from initialization and also initialize an empty `coupling_expression` we need slightly more QN-iterations in each time window than previous [reference results](https://github.com/precice/systemtests/blob/develop/tests/TestCompose_fe-fe/referenceOutput/precice-HeatDirichlet-iterations.log). This result is known and was observed during the design overhaul.

**New implementation**:
```
bcs = [DirichletBC(V, u_D, boundary)]
coupling_expression = precice.create_coupling_expression()
bcs.append(.., coupling_expression)

while precice._is_coupling_ongoing:
   ...
```

**Old implementation**:
```
initial_data = read_data from preCICE before coupling starts
bcs = [DirichletBC(V, u_D, boundary)]
coupling_expression = precice.create_coupling_expression(initial_data)
bcs.append(.., coupling_expression)

while precice._is_coupling_ongoing:
   ...
```

BenjaminRueth:
> ... we need slightly more QN-iterations in each time window than previous [reference results](https://github.com/precice/systemtests/blob/develop/tests/TestCompose_fe-fe/referenceOutput/precice-HeatDirichlet-iterations.log). 

The reason for this is probably (and hopefully) 3cd2fb8, where I intentionally removed initialization to see 1) whether our new scheme can handle cases without initialization and 2) what happens with the number of QN iterations. I would not expect a regression from this PR. I will now prepare a PR to update the tutorials according to this PR and carefully check for any regressions.

As soon as the tutorials PR is ready I would suggest to merge everything (don't forget to remove the `heat.py` from this PR again, since it was only introduced for prototyping).

BenjaminRueth:
I prepared https://github.com/precice/tutorials/pull/101 and checked that everything is working. I did not find anything that we have to worry about. So from my perspective this PR is ready to merge.

Issue #13, opened 2018-11-28
Small error at boundaries in fenics-fenics tutorial


BenjaminRueth:
Plotting the L2 error on the domain of the tutorial [HT/partitioned-heat/fenics-fenics](https://github.com/precice/tutorials/tree/master/HT/partitioned-heat/fenics-fenics) shows a small error at the coupling interface, if we are close to the boundary.

![smallerroratboundary](https://user-images.githubusercontent.com/5740604/49167496-c5abff80-f335-11e8-92e1-9b77fd43c766.png)

Inside the domain everything looks fine and the error also does not grow over time. If the resolution is reduced, the error gets smaller. These are generally good things. The error does not get smaller, if we reduce the tolerance of the coupling.

**What's wrong with this error?** We should be able to recover the exact solution in this example case:

* The analytical solution is first order in time and second order in space. Our discretization (implicit Euler/linear FEM) should be able to recover this solution. For details, see the [FEniCS tutorials book, p.40)[https://fenicsproject.org/tutorial/].
* The meshes are matching and therefore no mapping is necessary (which might introduce an error; see #11 )

**Where does this error come from?** Possible candidates:

* Wrong computation of the flux at the interface, since this is still a point where I am not 100% sure that we are doing everything correctly
* Wrong treatment of the point, where coupling interface and outer boundary meet.

The results from above can be recovered using https://github.com/precice/tutorials/commit/f9fe1201e1a9f648a2c325e04307d16bad5197e2 and https://github.com/precice/fenics-adapter/commit/a006b304325eabc9f15dba3d7c3b40d4a6a3b1bb.

BenjaminRueth:
If we use
```
         <relative-convergence-measure data="Flux"        mesh="NeumannNodes" limit="1e-10"/>
         <relative-convergence-measure data="Temperature" mesh="NeumannNodes" limit="1e-10"/>
```
in `precice-config.xml`, we can furtherly reduce the error.

![error](https://user-images.githubusercontent.com/5740604/49259104-4521f780-f438-11e8-9d17-9e01b8798472.png)

Furtherly reducing to `limit=1e-13` allows us to reach machine precision:

![error](https://user-images.githubusercontent.com/5740604/49259280-0fc9d980-f439-11e8-978d-c29d74d522a4.png)

uekerman:
We could then probably close this issue, right? It's a coupling error, no mapping error and no wrong boundary condition error. 
[...]

Issue #74, opened 2020-05-29
Make ExactInterpolationExpression safe


BenjaminRueth:
Our current implementation of `ExactInterpolationExpression` is limited to cases, where the coupling boundary is a straight line that is parallel to the y-axis. This is implicitly assumed by the way how we create our interpolant:

```
interp1d(self._coords_y, self._vals, bounds_error=False, fill_value="extrapolate", kind="cubic"))
```

As far as I know, there are no checks that prevent the user from using `ExactInterpolationExpression` on coupling boundaries that do not fullfill the requirements from above.

In my opinion it is sufficient to add security checks for now. With `GeneralInterpolationExpression` a good-enough alternative is available and `ExactInterpolationExpression` is only used for academic test cases, anyway.

BenjaminRueth:
Let's keep the discussion on the actual implementation of `ExactInterpolationExpression` that supports arbitrary geometry in #12. This issue is only about the security aspect of the current (restricted) implementation.

BenjaminRueth:
Not relevant anymore, since it was removed in #83.

Issue #12, opened 2018-11-29
Implement linear interpolation


BenjaminRueth:
originates from https://github.com/precice/fenics-adapter/issues/4#issuecomment-443031116.

BenjaminRueth:
Related to #10

BenjaminRueth:
This issue lacks some explanation. Let's try to provide this now:

Like explained in #74 `ExactInterpolationExpression` only supports coupling interfaces that are parallel to the y-axis. This restriction greatly simplifies the implementation, since we do not need connectivity information and we don't have to parametrize the coupling interface accordingly. However, for the general case, we have to provide this information. This general implementation is currently missing.

BenjaminRueth:
I see one problem with this issue:

We want to generalize our approach. Currently, we have the restriction that all edges of the coupling mesh **must** be parallel to the y-axis. This restriction allows us to just use the y-coordinate for parametrization of the interpolant. It also allows us to construct a higher order (currently [cubic](https://github.com/precice/fenics-adapter/blob/06c24b08fd5c475f85e623f1efff731d68eb9544/fenicsadapter/expression_core.py#L227-L228) interpolant and extrapolate).

The issue **Implement linear (or higher order?) interpolation** is twofold:

* linear interpolation for general case: No problem. Just use a linear interpolation along one edge.
* higher-order: Also requires data from neighboring elements/edges. This makes everything very complicated - especially if two edges form an angle.

However, if we restrict this issue to only supporting linear interpolation (!= cubic), this will also cost us some nice properties that the current approach offers: In the [heat-heat tutorial](https://github.com/precice/tutorials/blob/master/HT/partitioned-heat/fenics-fenics/heat.py) we know that the (manufactured) solution on the coupling interface must be quadratic. Therefore, we can recover the exact solution with a cubic interpolation. This allows us to use a very low error tolerance (which is nice for debugging). This whole setup breaks, if we do not support cubic interpolation anymore. But, as explained above, cubic interpolation cannot be offered for the general case.

My suggestion to solve this dilemma: Offer only linear interpolation (and an additional `LinearInterpolationExpression`) for the general case. Keep the current (non-general) state of the `ExactInterpolationExpression` for the sake of testing. I would also suggest to rename the issue to **Implement linear interpolation**, since I do not see a real solution for higher order.

BenjaminRueth:
A very simple approach for linear interpolation (with the tools that we already have) might be using `Rbf(x,y,function='linear')`:

```
[...]

PR #82, opened 2020-08-06
Change format of read data

BenjaminRueth:
This PR uses a `dict` instead of a `list` as format for `read_data`. Here the node position is used as keys and the former values of `read_data` are used as values. The new format allows the user to provide additional data or overwrite existing data. The implementation is backwards compatible and allows the user to improve the interpolation with known data, if required.

Another improvement: This feature makes extrapolation for `ExactInterpolationExpression` unnecessary, since we can just provide known data instead of relying on an extrapolation at the "gap" between the coupling interface and the physical boundary (see, for example https://github.com/precice/fenics-adapter/issues/13)

# Todo

- [x] update `PointSource` accordingly
- [ ] write test for `PointSource` @IshaanDesai 
- [x] merge https://github.com/precice/fenics-adapter/pull/81 (this PR is based on #81)
- [x] remove `prototyping` folder
- [x] merge #83 before this one

# Additional information

Originally I implemented this feature to increase the accuracy of the rbf interpolation (the accuracy is actually increased, but not sufficiently). I hoped to get rid of `ExactInterpolationExpression`. However, I realized that the problem with `GeneralInterpolationExpression` lies not only at the borders of the interpolant, but also between the nodes:

Figure_1

The example mimics the temperature at the coupling interface for [`heat.py`](https://github.com/precice/tutorials/blob/95cc1ed1294c9c9e5af4d6bcc6393be4cd8dcb53/HT/partitioned-heat/fenics-fenics/heat.py#L111-L112). 

The observed behaviour is related to the kind of rbf function that [`scipy.interpolate.Rbf`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.Rbf.html) uses. The `multiquadric` function just cannot represent a polynomial function exactly (while `ExactInterpolationExpression` is able to do this and therefore gives more accurate results). An alternative approach that might help us is provided in @uekerman's [thesis, section 4.3.2](https://mediatum.ub.tum.de/doc/1320661/1320661.pdf). However, this will require us to implement the interpolation on our own, as far as I understand it.


BenjaminRueth:
Related to #74, since extrapolation is unsafe.

uekerman:
There is an RBF alternative which should be easy to realize: to separate the polynomial from the RBF part. First, the polynomial part is fitted with the data, and afterwards the remaining error is interpolated only with the RBF part. So, the RBF interpolation of SciPy could be reused while still capturing the polynomial part exactly. I assume that there is also a least-squares fitting function in SciPy, basically a QR-decomposition.

A similar approach can also be used in preCICE. Mathematical details can be found in [this paper](ftp://ftp.informatik.uni-stuttgart.de/pub/library/ncstrl.ustuttgart_fi/INPROC-2017-35/INPROC-2017-35.pdf), page 6.

BenjaminRueth:
> 
> 
> There is an RBF alternative which should be easy to realize: to separate the polynomial from the RBF part. First, the polynomial part is fitted with the data, and afterwards the remaining error is interpolated only with the RBF part. So, the RBF interpolation of SciPy could be reused while still capturing the polynomial part exactly. I assume that there is also a least-squares fitting function in SciPy, basically a QR-decomposition.
> 
> A similar approach can also be used in preCICE. Mathematical details can be found in this paper, page 6: ftp://ftp.informatik.uni-stuttgart.de/pub/library/ncstrl.ustuttgart_fi/INPROC-2017-35/INPROC-2017-35.pdf .

fenicsadapter/fenicsadapter.py
```diff
@@ -109,21 +109,24 @@ def update_coupling_expression(self, coupling_expression, data):
...
-            for scalar valued data D = 1 and for vector valued data D = dimension of problem.
+        data : dict_like
+            The coupling data. A dictionary containing nodal data with vertex coordinates as key and associated data as
+            value.
         """
-        coupling_expression.update_boundary_data(data, self._coupling_mesh_vertices[:, 0], self._coupling_mesh_vertices[:, 1])
+        x_coord = np.array([vertex[0] for vertex in data.keys()])
+        y_coord = np.array([vertex[1] for vertex in data.keys()])
+        nodal_data = np.array([data for data in data.values()])
+        coupling_expression.update_boundary_data(nodal_data, x_coord, y_coord)
```
BenjaminRueth, line 119: I think we should also add some checks here (dimension fitting? coupling mesh vertices are still present in data?...)

BenjaminRueth:
First experiments with the segregated approach look promising:

segregated

Here is the code, in case anybody is interested:

```
import matplotlib.pyplot as plt
import numpy as np
from scipy.interpolate import Rbf
from scipy.linalg import lstsq

f = lambda x: 4 + 3*x**2
x = np.linspace(0,1,11)
y = np.random.rand(11)
y = f(x)
# a * x ** 2 + b * x + c ~= y
A = np.array([x**2, x, np.ones_like(x)]).T
w, _, _, _ = lstsq(A, y)
lstsq_interp = lambda x: w[0] * x**2 + w[1] * x + w[2]
res = y - lstsq_interp(x)
rbf_interp = Rbf(x,res)
interp = lambda x: rbf_interp(x) + lstsq_interp(x)
xx = np.linspace(0,1,51)
plt.title("segregated RBF interpolation of 4 + 3*x**2")
plt.plot(xx,f(xx),':',label="exact")
plt.plot(xx,interp(xx),label="interp")
plt.plot(x,y,'o',label="nodes")
plt.legend()
plt.show()
```

BenjaminRueth:
https://github.com/precice/fenics-adapter/pull/82/commits/433390bea371837bf2a5614c754bb2ecbb5be137 provides a working version using the segregated RBF interpolation. We can reach a similar level of accuracy as for cubic spline interpolation. Some notes on the current implementation:

* I only provided an implementation for 2d
* The results are slightly changing (coupling iterations etc), since the interpolation schemes (segregated RBF and cubic) work differently. However, the final result is the same within the given accuracy. -> **We will have to update our tests correspondingly.**
* Results for complex domain case are significantly improved (see pictures below)
* offering segregated RBF interpolation has the potential to close multiple issues:
    * #74, since we will not need `ExactInterpolationExpression` anymore
    * #13, since we generalize to the complex geometry case
    * #12, since I do not think we will need linear interpolation anymore

Note: I would suggest to still offer the "old" RBF interpolation, since the segregated version needs a minimum amount of data to create the fit and it is currently only provided for 2d cases.

# Complex case

## "old" RBF interpolation

complex_rbf

## segregated RBF interpolation

complex_rbf_seg

BenjaminRueth:
I decided to split up #82 into two PRs: This one for updating the `read_data` format (#82) and another one, which takes care of adding the new `SegregatedRBFInterpolationExpression` (link will be provided in a moment),

IshaanDesai:
Test for adapter function `get_point_sources` is temporarily disabled [here](https://github.com/precice/fenics-adapter/pull/82/commits/fc2c82d993644254cfd8900d81a670b9aa4637cf#diff-05e5a76b32a33c6ad22c8b60b4c7436fR190). 
The main concern is what do we test while checking the functionality for `get_point_sources`? More details in comment in code.

BenjaminRodenberg:
The results reported in https://github.com/precice/fenics-adapter/pull/82#issuecomment-672896370 cannot be reproduced with the current implementation of the fenics-adapter (https://github.com/precice/fenics-adapter/commit/300ab1a60770c72df0978556bd71e57476b67d69) and the tutorials (https://github.com/precice/tutorials/commit/107c076f2ed65a70cb0cd932387be0c19c0dc667). See #113 for details.

PR #86, opened 2020-10-01
Removes interpolation_type option and RBFInterpolationExpression

BenjaminRueth:
`SegregatedRBFInterpolationExpression` introduced in #83 can be used for all known use-cases of the adapter. This allows us to 

1. remove all other CouplingExpressions,
2. hard code the `Adapter._my_expression` and
3. remove the option `interpolation_type` from the configuration.

This simplifies the configuration and the code. Note that we still keep the object-oriented design, where `CouplingExpression` may be used as a basis for new coupling expressions.

### DIFF
--- a/fenicsadapter/config.py
+++ b/fenicsadapter/config.py
@@ -46,10 +46,6 @@ class Config:
         except:
             self._write_data_name = None
         self._read_data_name = data["interface"]["read_data_name"]
-        try:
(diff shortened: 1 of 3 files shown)


IshaanDesai (approved):
Looks good! The whole Expression mechanism looks cleaner now that we have only one interpolation strategy :+1:

PR #34, opened 2019-04-29
WIP: Refactor control flow

BenjaminRueth:
Working on #33 in this PR

PR #87, opened 2020-10-02
Add capability to write delta change data

IshaanDesai:
This PR adds the functionality to existing data writing function of writing the delta change of a quantity between the current time window and the previous time window. Previously the data writing function only supported writing absolute value of a quantity at the current time window. 
In coupled FSI applications this is typically seen as the sturcutral participant writing `Displacements` or `DisplacementDeltas`.

### DIFF
--- a/fenicsadapter/config.py
+++ b/fenicsadapter/config.py
@@ -20,7 +20,7 @@ class Config:
         self._coupling_mesh_name = None
         self._read_data_name = None
         self._write_data_name = None
-        self._interpolation_type = None
+        self._data_writing_type = None
 
         self.read_json(adapter_config_filename)
 
@@ -46,6 +46,7 @@ class Config:
         except:
             self._write_data_name = None
         self._read_data_name = data["interface"]["read_data_name"]
+        self._data_writing_type = data["interface"]["data_writing_type"]
 
         read_file.close()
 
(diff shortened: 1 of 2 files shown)


fenicsadapter/config.py
```diff
@@ -20,7 +20,7 @@ def __init__(self, adapter_config_filename):
         self._coupling_mesh_name = None
         self._read_data_name = None
         self._write_data_name = None
-        self._interpolation_type = None
```
BenjaminRueth, line 23: I think this line should already have been removed in #86. Changing it now to `self._data_writing_type` gives a wrong impression. I already created another PR #88 for cleaning this up.

fenicsadapter/config.py
```diff
@@ -46,6 +46,7 @@ def read_json(self, adapter_config_filename):
         except:
             self._write_data_name = None
         self._read_data_name = data["interface"]["read_data_name"]
+        self._data_writing_type = data["interface"]["data_writing_type"]
```
BenjaminRueth, line 49: I think it is good that we make the `data_writing_type` configurable. What does it default to? Please make sure that it is an optional argument in our configuration file.

IshaanDesai: I think the default should be `"absolute"`, so by default the adapter will write absolute values of the write data. Is that okay?

BenjaminRueth: Yes. Totally agree. Please make sure to also document the meaning of these two keywords somewhere.

BenjaminRueth:
We just decided to let the user do these kind of calculations, since our design goal with the adapter is to convert and communicate the data, but not touch it (#34) and getting the "displacement delta" from "displacement" is easy. Closing this issue.

PR #92, opened 2020-10-22
Use github actions for unit tests

### LINKED ISSUE #91: Use Github Actions CI
preCICE CI was moved to Github Actions in https://github.com/precice/precice/pull/843. We should also move our testing infrastructure for the fenics-adapter to Github Actions.

BenjaminRueth:
Closes #91 

- [x] do unit tests using Github Actions


BenjaminRueth:
@fsimonis can you check whether the setup of the workflow is in compliance with what we do for `precice/precice`. Please note that there currently seems to be a problem with FEniCS installation using `apt` (https://fenicsproject.discourse.group/t/ubuntu-18-04-importerror-cannot-import-name-sub-forms-by-domain/4257). I could not solve this problem and therefore I'm not using [`precice/ci-ubuntu-1804`](https://hub.docker.com/r/precice/ci-ubuntu-1804) or [`precice/ci-ubuntu-2004`](https://hub.docker.com/r/precice/ci-ubuntu-2004) and installing FEniCS on top.

Instead, I'm using `quay.io/fenicsproject/stable:current`, which is the official container from FEniCS (https://fenicsproject.org/download/) that seems to work. If necessary, we can put preCICE on top of that container (or hope that the problem with the FEniCS installation from `apt` will get fixed soon).

Eder-K:
I think it's fine to keep it the Travis command for now, at the moment the migration of systemtests is still being planned and not very far along.

BenjaminRueth:
> Even though it's a little ugly this way, I think it's fine to keep it the Travis command for now, at the moment the migration of systemtests is still being planned and not very far along.

Thanks for the feedback @Eder-K! Then this PR is from my perspective ready for review & merge.

Eder-K (approved):

BenjaminRueth:
Background information: We are using `quay.io/fenicsproject/stable:current` and not one of "our" preCICE containers, like in https://github.com/precice/systemtests/blob/develop/adapters/Dockerfile.fenics-adapter, since currently the FEniCS installation via apt has two bigger problems in the context of preCICE:

1. Installing PETSc from source breaks FEniCS: https://fenicsproject.discourse.group/t/modulenotfounderror-no-module-named-dolfin-if-petsc-dir-is-set/4407
2. Installing a python package with FEniCS as dependency breaks FEniCS: https://fenicsproject.discourse.group/t/installing-python-package-with-fenics-dependency-breaks-fenics-installation/4476

Both issues can be avoided, but in this context the easiest solution is to just use `quay.io/fenicsproject/stable:current`.

PR #95, opened 2020-11-02
Add CHANGELOG.md

BenjaminRueth:
Now that we are moving closer to a v1.0.0 release, we should also provide a Changelog. 

@IshaanDesai and @uekerman any important changes, features that we should include in the Changelog? What should we do about features that were introduced much earlier? Example: The major design overhaul of the API looks to me a bit funny, if the API is never mentioned in the changelog before.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,13 @@
 
 ## 1.0.0
 
-* Use naming proposed in https://github.com/precice/fenics-adapter/issues/85. The software is called FEniCS-preCICE adapter, the corresponding python package `fenicsprecice`
-* Deprecate old package name `fenicsadapter`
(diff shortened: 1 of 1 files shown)


uekerman:
That's a bit a philosophical question :smile: 

In theory to start with a changelog only makes sense after a release, right?
From a user perspective we explain the functionality of the adapter in our paper. Afterwards the changelog can explain what we gradually change from there on. In the end the changelog is where you learn about what changes from version to version, not where you learn about the basics of a software.

I see two options:
* As a first entry we could mention the main features of the adapter and refer to the paper. This should be very short, we should not duplicate the paper.
* We leave the v1 entry completely blank.

BenjaminRueth:
I changed the wording to describe the "current state" rather than the "diff". What do you think?

uekerman (approved):
Looks good now.

IshaanDesai (approved):

Issue #89, opened 2020-10-06
Provide robust initialization 


IshaanDesai:
Currently the user has to provide `coupling_subdomain`, `mesh` and `function_space` as mandatory arguments of `initialize`. `write_function=None`, `fixed_boundary=None` are optional arguments.

In a general FEniCS setup the read function and write function need **not** live on the same function space `V` (see [`heat.py`](https://github.com/precice/tutorials/blob/95cc1ed1294c9c9e5af4d6bcc6393be4cd8dcb53/HT/partitioned-heat/fenics-fenics/heat.py#L107-L108)). The current design of the adapter assumes that the user provides the function space associated with the read function to initialize (see [`heat.py`](https://github.com/precice/tutorials/blob/95cc1ed1294c9c9e5af4d6bcc6393be4cd8dcb53/HT/partitioned-heat/fenics-fenics/heat.py#L125-L132)). This prompts the need to extract additional information of the function space associated with the write function when writing data to preCICE.

A more general initialization routine can handle this smartly. The way to approach this is to use this adapter in more and more advanced FEniCS applications and modify the design by custom requirement. Parallelization of the adapter in https://github.com/precice/fenics-adapter/pull/71 already highlights such a design methodology.

Currently the same initialization routine is used if a user wants to read and write data via FEniCS Function / Expression objects or if the user wants to handle multiple `PointSource` objects. Further design ideas to differentiate these two cases need to be implemented.

BenjaminRueth:
> This prompts the need to extract additional information of the function space associated with the write function when writing data to preCICE.

Which kind of information do you exactly mean here?

> The way to approach this is to use this adapter in more and more advanced FEniCS applications and modify the design by custom requirement.

I think we should try to make this more concrete. How do you imagine the new signature of `initialize` that fulfils all requirements? Can we already estimate what the requirements are or are there still too many unknowns? The only "advanced" example I have in mind is (of course) the heat transport tutorial. I edited your original post above and added some examples, if I was able to find good ones.

PR #101, opened 2020-11-04
Overhaul Initialization

IshaanDesai:
This PR overhauls the initialization to make it more versatile. The new initialization function can be used to initialize uni-directional coupling cases. The proposed initialization also addresses: https://github.com/precice/fenics-adapter/issues/89 and makes the adapter initialization more robust.
Closes https://github.com/precice/fenics-adapter/issues/89
To-Do before merging this:
- [x] Update tutorial cases to handle the new signature of `initialize()` (https://github.com/precice/tutorials/pull/119)
- [x] Update mock tests

### DIFF
--- a/fenicsprecice/fenicsprecice.py
+++ b/fenicsprecice/fenicsprecice.py
@@ -263,19 +263,21 @@ class Adapter:
 
     def initialize(self, coupling_subdomain, read_function_space=None, write_object=None, fixed_boundary=None):
         """
-        Initializes the coupling interface and sets up the mesh in preCICE. Custom initialization also possible.
+        Initializes the coupling interface and sets up the mesh in preCICE.
 
         Parameters
         ----------
         coupling_subdomain : Object of class dolfin.cpp.mesh.SubDomain
             SubDomain of mesh which is the physical coupling boundary.
         read_function_space : Object of class dolfin.functions.functionspace.FunctionSpace
-            Function space on which the read function lives.
+            Function space on which the read function lives. If not provided then the adapter assumes that this
+            participant is a write-only participant.
         write_object : Object of class dolfin.functions.functionspace.FunctionSpace OR dolfin.functions.function.Function
             Function space on which the write function lives or FEniCS function related to the quantity to be written
-            by FEniCS during each coupling iteration.
+            by FEniCS during each coupling iteration. If not provided then the adapter assumes that this participant is
+            a read-only participant.
         fixed_boundary : Object of class dolfin.fem.bcs.AutoSubDomain
-            SubDomain consisting of a fixed boundary condition. For example in FSI cases usually the solid body
+            SubDomain consisting of a fixed boundary of the mesh. For example in FSI cases usually the solid body
             is fixed at one end (fixed end of a flexible beam).
 
         Returns
--- a/fenicsprecice/fenicsprecice.py
+++ b/fenicsprecice/fenicsprecice.py
@@ -310,7 +310,8 @@ class Adapter:
         _, self._fenics_dimensions = coords.shape
 
         # Ensure that function spaces of read and write functions use the same mesh
-        assert self._read_function_space.mesh() is self._write_function_space.mesh()
+        if self._coupling_type is CouplingMode.BIDIR_COUPLING:
+            assert self._read_function_space.mesh() is self._write_function_space.mesh()
 
         if fixed_boundary:
             self._Dirichlet_Boundary = fixed_boundary
--- a/fenicsprecice/fenicsprecice.py
+++ b/fenicsprecice/fenicsprecice.py
@@ -301,6 +301,13 @@ class Adapter:
             raise Exception("Given write object is neither of type dolfin.functions.function.Function or "
                             "dolfin.functions.functionspace.FunctionSpace")
 
+        if type(read_function_space) is FunctionSpace:
+            pass
+        elif read_function_space is None:
+            pass
+        else:
+            raise Exception("Given read_function_space is not of type dolfin.functions.functionspace.FunctionSpace")
+
         if read_function_space is None and write_function_space:
             self._coupling_type = CouplingMode.UNIDIR_WRITE
             assert (self._config.get_write_data_name())
--- a/fenicsprecice/fenicsprecice.py
+++ b/fenicsprecice/fenicsprecice.py
@@ -9,6 +9,7 @@ from .adapter_core import FunctionType, CouplingMode, determine_function_type, c
     get_coupling_boundary_vertices, get_coupling_boundary_edges, get_forces_as_point_sources
 from .expression_core import SegregatedRBFInterpolationExpression
 from .solverstate import SolverState
+from fenics import Function, FunctionSpace
 from warnings import warn
 
 logger = logging.getLogger(__name__)
@@ -48,7 +49,6 @@ class Adapter:
         # FEniCS related quantities
         self._fenics_dimensions = None
         self._read_function_space = None  # initialized later
-        self._write_function_space = None  # initialized later
 
         # coupling mesh related quantities
         self._coupling_mesh_vertices = None  # initialized later
@@ -262,9 +262,9 @@ class Adapter:
         else:
             raise Exception("write_function provided is neither VECTOR nor SCALAR type")
 
-    def initialize(self, coupling_subdomain, read_function_space=None, write_function_space=None, write_function=None, fixed_boundary=None):
+    def initialize(self, coupling_subdomain, read_function_space=None, write_object=None, fixed_boundary=None):
         """
-        Initializes the coupling interface and sets up the mesh in preCICE. Allows to initialize data on coupling interface.
+        Initializes the coupling interface and sets up the mesh in preCICE. Custom initialization also possible.
 
         Parameters
         ----------
@@ -272,10 +272,9 @@ class Adapter:
             SubDomain of mesh which is the physical coupling boundary.
         read_function_space : Object of class dolfin.functions.functionspace.FunctionSpace
             Function space on which the read function lives.
-        write_function_space : Object of class dolfin.functions.functionspace.FunctionSpace
-            Function space on which the write function lives.
-        write_function : Object of class dolfin.functions.function.Function
-            FEniCS function related to the quantity to be written by FEniCS during each coupling iteration.
+        write_object : Object of class dolfin.functions.functionspace.FunctionSpace OR dolfin.functions.function.Function
+            Function space on which the write function lives or FEniCS function related to the quantity to be written
+            by FEniCS during each coupling iteration.
         fixed_boundary : Object of class dolfin.fem.bcs.AutoSubDomain
             SubDomain consisting of a fixed boundary condition. For example in FSI cases usually the solid body
             is fixed at one end (fixed end of a flexible beam).
@@ -285,6 +284,16 @@ class Adapter:
         dt : double
             Recommended time step value from preCICE.
         """
+
+        if type(write_object) is Function:
+            write_function_space = write_object.function_space()
+            write_function = write_object
+        elif type(write_object) is FunctionSpace:
+            write_function_space = write_object
+            write_function = None
+        else:
+            raise Exception("Given write object is neither of type dolfin.functions.function.Function or "
+                            "dolfin.functions.functionspace.FunctionSpace")
 
         if read_function_space is None and write_function_space:
             self._coupling_type = CouplingMode.UNIDIR_WRITE
@@ -305,14 +314,13 @@ class Adapter:
 
         if self._coupling_type is CouplingMode.UNIDIR_WRITE or self._coupling_type is CouplingMode.BIDIR:
             self._write_function_type = determine_function_type(write_function_space)
-            self._write_function_space = write_function_space
 
         coords = self._read_function_space.tabulate_dof_coordinates()
         _, self._fenics_dimensions = coords.shape
 
         # Ensure that function spaces of read and write functions use the same mesh
         if self._coupling_type is CouplingMode.BIDIR:
-            assert(self._read_function_space.mesh() is self._write_function_space.mesh())
+            assert(self._read_function_space.mesh() is write_function_space.mesh())
 
         if fixed_boundary:
             self._Dirichlet_Boundary = fixed_boundary
--- a/fenicsprecice/fenicsprecice.py
+++ b/fenicsprecice/fenicsprecice.py
@@ -5,7 +5,7 @@ import numpy as np
 from .config import Config
 import logging
 import precice
-from .adapter_core import FunctionType, determine_function_type, convert_fenics_to_precice, \
+from .adapter_core import FunctionType, CouplingMode, determine_function_type, convert_fenics_to_precice, \
     get_coupling_boundary_vertices, get_coupling_boundary_edges, get_forces_as_point_sources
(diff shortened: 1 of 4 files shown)


BenjaminRueth (changes requested):
Looking at this PR and #102 I would suggest to first merge this one and then #102. Some comments of my latest review on #102 also apply here. I still have to do a thorough review on the rest of this PR, though.

IshaanDesai:
> Looking at this PR and #102 I would suggest to first merge this one and then #102. Some comments of my latest review on #102 also apply here. I still have to do a thorough review on the rest of this PR, though.

This PR is ready for review alongside https://github.com/precice/tutorials/pull/119

BenjaminRueth:
As reference for the future and addiitonal information for the related PR (e.g. https://github.com/precice/tutorials/pull/119#discussion_r530183225) I will quickly try to summarize all the alternatives that I and @IshaanDesai have also discussed:

As a general remark: `argument=None` means that the argument is optional.

### Old initialization

 `def initialize(self, coupling_subdomain, mesh, function_space, write_function=None, fixed_boundary=None)`

**Problems:** 

1. Unclear: Does `function_space` mean read, write or both function spaced?
2. What if `write_function` uses a different function space than `function_space`?
3. Does not support one-way-coupling, since `function_space` is a must.
4. Does not support different read and write function spaces (if we assume that this space is automatically used for both sides).

### Read function space + write function

 `def initialize(self, coupling_subdomain, read_function_space=None, write_function=None, fixed_boundary=None)`

**Discussion of problems from above + new ones**

1. Explicitly states whether read or write function space.
2. We directly use `write_function.function_space()`, if we have to access it. No problem.
3. Supports one-way-coupling.
4. Supports different read and write function spaces (`read_function_space` and `write_function.function_space()`).
5. **new** User **must** provide `write_function` for two-way-coupling, even, if user can only provide `write_function_space` at initialization. Image use-case: User does not want to / cannot `initialize_data`. Possible workaround: Just initialize zero function on `write_function_space`, but this is ugly.

### Two function spaces + write function

 `def initialize(self, coupling_subdomain, read_function_space=None, write_function_space=None, write_function=None, fixed_boundary=None)`

**Discussion of problems from above + new ones**

1. Explicitly states whether read or write function space.
2. Assertion ensures that `write_function_space == write_function.function_space()`, if both are provided.
3. Supports one-way-coupling.
4. Supports different read and write function spaces.
5. User can just provide `write_function_space` instead of `write_function`.
6. **new** Arguments `write_function_space` and `write_function` are to some degree redundant. We can avoid mistakes through an assertion (see 2.), but still the API is not perfect.

### Initialization in this PR

 `def initialize(self, coupling_subdomain, read_function_space=None, write_object=None, fixed_boundary=None)`

**Discussion of problems from above + new ones**

1. Explicitly states whether read or write function space/function (`write_object`).
2. No problem.
3. Supports one-way-coupling.
4. Supports different read and write function spaces.
5. User can just provide a `write_function_space` or a `write_function` as `write_object`.
6. Only `write_function_space` or `write_function` may be provided to `write_object`. No need for assertions to ensure consistency.
7. **new:** see https://github.com/precice/tutorials/pull/119#discussion_r530183225. If we have identical function spaces, we pass `V` twice (problem already existed for most approaches above, as well).

@IshaanDesai: If I'm missing something feel free to edit this comment.

IshaanDesai:
Documentation of discussion looks good. System test will fail till we merge the corresponding tutorials PR. This PR is ready for merging

fenicsprecice/fenicsprecice.py
```diff
@@ -273,9 +286,47 @@ def initialize(self, coupling_subdomain, mesh, function_space, write_function=No
...
         """
 
-        coords = function_space.tabulate_dof_coordinates()
+        write_function_space, write_function = None, None
+        if type(write_object) is Function:
+            write_function_space = write_object.function_space()
+            write_function = write_object
+        elif type(write_object) is FunctionSpace:
+            write_function_space = write_object
+            write_function = None
```
BenjaminRueth, line 295:
```suggestion
        if type(write_object) is Function:  # precice.initialize_data() will be called, since fenics.Function is provided by user
            write_function_space = write_object.function_space()
            write_function = write_object
        elif type(write_object) is FunctionSpace:  # only a FunctionSpace is provided by user. preCICE will use default zero values for initialization.
            write_function_space = write_object
            write_function = None
```
I think here a comment would be helpful in order to document what thes two branches actually mean:

fenicsprecice/fenicsprecice.py
```diff
@@ -273,9 +286,47 @@ def initialize(self, coupling_subdomain, mesh, function_space, write_function=No
...
+        elif write_object is None:
+            print("Participant {} is read-only participant".format(self._config.get_participant_name()))
+        else:
+            raise Exception("Given write object is neither of type dolfin.functions.function.Function or "
+                            "dolfin.functions.functionspace.FunctionSpace")
```
BenjaminRueth, line 300:
```suggestion
            raise Exception("Given write_object is neither of type dolfin.functions.function.Function or "
                            "dolfin.functions.functionspace.FunctionSpace. Please provide a Function, if you "
                            "want to initialize coupling data (via precice.initialize_data()) according to this "
                            "function. A FunctionSpace is sufficient, if you don't want to initialize data. If you 
                            "want to do one-way coupling where this participant is only reading data, just "
                            "leave the write_object blank.")
```
Might be a bit verbose, but then the user really knows what's going on.

fenicsprecice/fenicsprecice.py
```diff
@@ -273,9 +286,47 @@ def initialize(self, coupling_subdomain, mesh, function_space, write_function=No
...
+
+        if read_function_space is None and write_function_space:
+            self._coupling_type = CouplingMode.UNIDIR_WRITE
+            assert(self._config.get_write_data_name())
+            print("Participant {} is write-only participant".format(self._config.get_participant_name()))
+        elif read_function_space and write_function_space is None:
+            self._coupling_type = CouplingMode.UNIDIR_READ
+            assert(self._config.get_read_data_name())
```
BenjaminRueth, line 308: I would suggest to put the user feedback from above (https://github.com/precice/fenics-adapter/pull/101/files#diff-6837aaf707ece32d403dde56a219c19c3f3ddf26c753a20a3ac4259a4f7fe4aaR297) to this place. Then the code is easier to read. In the section above you can simply:
```
elif write_object is None:
    pass
````
We will also avoid unnecessary output, if `read_function_space is None and write_function_space is None`. Currently, you will first `print(print("Participant {} is read-only participant".format(self._config.get_participant_name())))`, if `write_function_space is None` (https://github.com/precice/fenics-adapter/pull/101/files#diff-6837aaf707ece32d403dde56a219c19c3f3ddf26c753a20a3ac4259a4f7fe4aaR296-R297). However, a few lines later we enter the `else` and `raise Exception(...)` (https://github.com/precice/fenics-adapter/pull/101/files#diff-6837aaf707ece32d403dde56a219c19c3f3ddf26c753a20a3ac4259a4f7fe4aaR312-R314).

File: fenicsprecice/fenicsprecice.py [hunk 4]
```diff
@@ -273,9 +286,47 @@ def initialize(self, coupling_subdomain, mesh, function_space, write_function=No
...
+        elif read_function_space and write_function_space:
+            self._coupling_type = CouplingMode.BIDIR
+            assert(self._config.get_read_data_name() and self._config.get_write_data_name())
+        else:
+            raise Exception("Incorrect read and write function space combination provided. Please check input "
+                            "parameters in initialization")
```
BenjaminRueth, line 314:
```suggestion
        elif read_function_space is None and write_function_space is None:
            raise Exception("You did not provide a read_function_space or write_function_space. "
                                      "Please provide a write_function_space, for one-way coupling, where this "
                                      "participant only writes. Please provide a read_function_space, for one-way "
                                      "coupling, where this participant only reads. If you want to realize two-way"
                                      "coupling provide both, read_function_space and write_function_space.")            
        else:
            raise Exception("Incorrect read and write function space combination provided. Please check input "
                            "parameters in initialization")
```
Again, might be a bit verbose, but we should check for this specific case and give a specific error. The `else:` is for everything that we do not expect.

fenicsprecice/fenicsprecice.py
```diff
@@ -294,8 +345,8 @@ def initialize(self, coupling_subdomain, mesh, function_space, write_function=No
                                 "No proper treatment for dimensional mismatch is implemented. Aborting!".format(
                     self._fenics_dimensions, self._interface.get_dimensions()))
 
-        fenics_vertices, self._coupling_mesh_vertices = get_coupling_boundary_vertices(
-            mesh, coupling_subdomain, self._fenics_dimensions, self._interface.get_dimensions())
+        fenics_vertices, self._coupling_mesh_vertices = get_coupling_boundary_vertices(self._read_function_space.mesh(),
+                coupling_subdomain, self._fenics_dimensions, self._interface.get_dimensions())
```
BenjaminRueth, line 349: What happens, `if read_function_space is None`? I would rather simply use a `mesh` here that has been initialized previously according to the `CouplingMode`.

BenjaminRueth: I think this also improves readability.

fenicsprecice/fenicsprecice.py
```diff
@@ -307,18 +358,15 @@ def initialize(self, coupling_subdomain, mesh, function_space, write_function=No
         for i in range(n_vertices):
             id_mapping[fenics_vertices[i].global_index()] = self._vertex_ids[i]
 
-        edge_vertex_ids1, edge_vertex_ids2 = get_coupling_boundary_edges(mesh, coupling_subdomain, id_mapping)
+        edge_vertex_ids1, edge_vertex_ids2 = get_coupling_boundary_edges(self._read_function_space.mesh(),
+                                                                         coupling_subdomain, id_mapping)
```
BenjaminRueth, line 362: Same like above. Better simply use `mesh`.

fenicsprecice/fenicsprecice.py
```diff
@@ -273,9 +286,47 @@ def initialize(self, coupling_subdomain, mesh, function_space, write_function=No
...
+
+        if self._coupling_type is CouplingMode.UNIDIR_WRITE or self._coupling_type is CouplingMode.BIDIR:
+            self._write_function_type = determine_function_type(write_function_space)
+
+        coords = self._read_function_space.tabulate_dof_coordinates()
         _, self._fenics_dimensions = coords.shape
 
+        # Ensure that function spaces of read and write functions use the same mesh
+        if self._coupling_type is CouplingMode.BIDIR:
+            assert(self._read_function_space.mesh() is write_function_space.mesh())
```
BenjaminRueth, line 328:
```suggestion
        if self._coupling_type is CouplingMode.BIDIR:
            # Ensure that function spaces of read and write functions use the same mesh
            assert(self._read_function_space.mesh() is write_function_space.mesh()), "read_function_space and write_function_space must use the same mesh."
            mesh = self._read_function_space.mesh()
        elif self._coupling_type is CouplingMode.UNIDIR_READ:
            mesh = self._read_function_space.mesh()
        elif self._coupling_type is CouplingMode.UNIDIR_WRITE:
            mesh = self._write_function_space.mesh()
        else:
            raise Exception("Unexpected control flow.")
```
Please add some meaningful error message after the assertion.

BenjaminRueth (changes requested):
Generally the changes look good. All my comments are mainly on user interaction and code security aspects.

If there are no major points on the discussion https://github.com/precice/fenics-adapter/pull/101#issuecomment-733591919, I would say let's merge this PR as soon as my comments are resolved.

fenicsprecice/fenicsprecice.py
```diff
@@ -273,9 +288,60 @@ def initialize(self, coupling_subdomain, mesh, function_space, write_function=No
...
+
+        if self._coupling_type is CouplingMode.UNIDIR_READ or self._coupling_type is CouplingMode.BIDIR:
+            self._read_function_type = determine_function_type(read_function_space)
+            self._read_function_space = read_function_space
+
+        if self._coupling_type is CouplingMode.UNIDIR_WRITE or self._coupling_type is CouplingMode.BIDIR:
+            # Ensure that function spaces of read and write functions are defined using the same mesh
+            self._write_function_type = determine_function_type(write_function_space)
+
+        coords = self._read_function_space.tabulate_dof_coordinates()
```
BenjaminRueth, line 336: I just ran into the following error:

```
Participant HeatNeumann is write-only participant
Traceback (most recent call last):
  File "heat.py", line 130, in <module>
    precice_dt = precice.initialize(coupling_boundary, write_object=u_D_function)
  File "/home/benjamin/.local/lib/python3.8/site-packages/fenicsprecice/fenicsprecice.py", line 336, in initialize
    coords = self._read_function_space.tabulate_dof_coordinates()
AttributeError: 'NoneType' object has no attribute 'tabulate_dof_coordinates'
---[precice]  Implicitly finalizing in destructor
```

Used the following modified version of the HT tutorial: 
[l_336_error.txt](https://github.com/precice/fenics-adapter/files/5607978/l_336_error.txt)

Note: you can use `git apply l_336_error.txt` to apply the provided changes.

fenicsprecice/fenicsprecice.py
```diff
@@ -273,9 +288,60 @@ def initialize(self, coupling_subdomain, mesh, function_space, write_function=No
...
+        elif read_function_space and write_function_space:
+            self._coupling_type = CouplingMode.BIDIR
+            assert (self._config.get_read_data_name() and self._config.get_write_data_name())
+            mesh = read_function_space.mesh()
+        elif read_function_space is None and write_function_space is None:
+            raise Exception("Neither read_function_space nor write_function_space is provided. Please provide a write "
+                            "function_space if one-way coupling with this participant only writing data is intended. "
+                            "Please provide a read_function_space if one-way coupling with this participant only "
+                            "reading data is intended. If two-way coupling is implemented then both read_function_space"
+                            " and write_object need to be provided.")
```
BenjaminRueth, line 323:
```suggestion
            raise Exception("Neither read_function_space nor write_function_space is provided. Please provide a "
                            "write_object if this participant is used in one-way coupling and only writes data. "
                            "Please provide a read_function_space if this participant is used in one-way coupling and "
                            "only reads data. If two-way coupling is implemented then both read_function_space"
                            " and write_object need to be provided.")
```
This exception is triggered correctly. But I realized that what I have suggested before is almost unreadable... Do you think this message is an improvement?

fenicsprecice/fenicsprecice.py
```diff
@@ -273,9 +288,60 @@ def initialize(self, coupling_subdomain, mesh, function_space, write_function=No
...
+        else:
+            raise Exception("Incorrect read and write function space combination provided. Please check input "
+                            "parameters in initialization")
```
BenjaminRueth, line 326: I tried to break the code and provided `read_function_space=dict()`. This brought me to this exception. Why do we check for the correct type of the `write_object` above, but not for the correct type of the `read_function_space`? I think it would be reasonable to also do this already above, before we start checking the different possible combinations.

BenjaminRueth (changes requested):
I tried to run a one-way-coupling case, but ran into a problem. I commented on this and some other minor things that I found when trying to break the code below.

Do you think it makes sense to also provide a case with one-way coupling? Should be rather easy now and might be helpful, if we add this to our tutorials (HT would be enough). As far as I understand we only have to modify the initialization and the adapter config (do we even have to do this?).

PR #102, opened 2020-11-05
Remove pseudo 2D-3D coupling

IshaanDesai:
The strategy to handle pseudo 2D-3D cases is now changed in the following manner: 
* **Previous approach**: The 3D participant solves a full 3D simulation with one direction having only one mesh element. The 2D participant reads 3D data from preCICE and ignore the direction which had the single mesh element. 
* **New approach**: The 3D participant solves a 3D case but writes only 2D data to preCICE. The 2D participant reads 2D data and uses it directly without any need for modification.

This is already discussed with reference to tutorials here: https://github.com/precice/tutorials/issues/112. In light of the new approach the `fenics-adapter` no longer needs to handle pseudo 2D-3D coupling. This PR removes pseudo 2D-3D coupling capability in the `fenics-adapter`.

### DIFF
--- a/fenicsprecice/adapter_core.py
+++ b/fenicsprecice/adapter_core.py
@@ -157,19 +157,13 @@ def get_coupling_boundary_vertices(mesh_fenics, coupling_subdomain, fenics_dimen
             vertices_x.append(v.x(0))
             if dimensions == 2:
                 vertices_y.append(v.x(1))
-            elif fenics_dimensions == 2 and dimensions == 3:
-                vertices_y.append(v.x(1))
-                vertices_z.append(0)
             else:
                 raise Exception("Dimensions of coupling problem (dim={}) and FEniCS setup (dim={}) do not match!"
                                 .format(dimensions, fenics_dimensions))
 
     assert (n != 0), "No coupling boundary vertices detected"
 
-    if dimensions == 2:
-        return fenics_vertices, np.stack([vertices_x, vertices_y], axis=1)
-    elif dimensions == 3:
-        return fenics_vertices, np.stack([vertices_x, vertices_y, vertices_z], axis=1)
+    return fenics_vertices, np.stack([vertices_x, vertices_y], axis=1)
 
 
 def are_connected_by_edge(v1, v2):
@@ -241,7 +235,7 @@ def get_coupling_boundary_edges(mesh_fenics, coupling_subdomain, id_mapping):
     return vertices1_ids, vertices2_ids
 
 
-def get_forces_as_point_sources(fixed_boundary, function_space, coupling_mesh_vertices, data, z_dead=False):
+def get_forces_as_point_sources(fixed_boundary, function_space, coupling_mesh_vertices, data):
     """
     Creating two dicts of PointSources that can be applied to the assembled system. Appling filter_point_source to
     avoid forces being applied to already existing Dirichlet BC, since this would lead to an overdetermined system
@@ -259,8 +253,6 @@ def get_forces_as_point_sources(fixed_boundary, function_space, coupling_mesh_ve
         numpy array [N x D] where N = number of vertices and D = dimensions of geometry
     data : PointSource
         FEniCS PointSource data carrying forces
-    z_dead: bool
-        Allows to ignore z dimension
 
     Returns
     -------
@@ -278,12 +270,6 @@ def get_forces_as_point_sources(fixed_boundary, function_space, coupling_mesh_ve
 
     # Check for shape of coupling_mesh_vertices and raise Assertion for 3D
     n_vertices, dims = coupling_mesh_vertices.shape
-
-    if z_dead:
-        assert (dims == 3), "z_dead=True is only allowed for 3D data"
-    else:
-        assert (dims == 2), "This Adapter can create Point Sources only from 2D data. Use z_dead=True, if you want " \
-                            "to ignore the z dimension."
 
     vertices_x = coupling_mesh_vertices[:, 0]
     vertices_y = coupling_mesh_vertices[:, 1]
--- a/fenicsprecice/adapter_core.py
+++ b/fenicsprecice/adapter_core.py
@@ -145,7 +145,6 @@ def get_coupling_boundary_vertices(mesh_fenics, coupling_subdomain, fenics_dimen
     fenics_vertices = []
     vertices_x = []
     vertices_y = []
-    vertices_z = []
 
     if not issubclass(type(coupling_subdomain), SubDomain):
         raise Exception("No correct coupling interface defined! Given coupling domain is not of type dolfin Subdomain")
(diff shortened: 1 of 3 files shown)


IshaanDesai:
Proper handling of 1D cases also needs to be done. @BenjaminRueth is it okay to do in this same PR?

uekerman:
> Proper handling of 1D cases also needs to be done.

To what kind of 1D cases are you referring?  Just to avoid misunderstandings: preCICE cannot handle 1D cases (yet). So you cannot specify `dimensions="1"` in the preCICE config.

IshaanDesai:
> To what kind of 1D cases are you referring? Just to avoid misunderstandings: preCICE cannot handle 1D cases (yet). So you cannot specify `dimensions="1"` in the preCICE config.

Okay then my comment is not relevant. I was assuming that 1D case definition is possible in preCICE.

fenicsprecice/adapter_core.py
```diff
@@ -22,6 +22,17 @@ class FunctionType(Enum):
...
 
+class CouplingMode(Enum):
+    """
+    Defines the type of coupling being used.
+    Options are: Bi-directional coupling, Uni-directional Write Coupling, Uni-directional Read Coupling
+    Used in assertions to check which type of coupling is done
+    """
+    BIDIR = 4
+    UNIDIR_WRITE = 5
+    UNIDIR_READ = 6
```
BenjaminRueth, line 33: I would suggest to write the full names here. `BIDIRECTIONAL_COUPLING` or `TWO_WAY_COUPLING`. Makes it easier to read. Btw: Why do you start counting at `4`? I think that one `Enum` does not influence the other one.

Additionally: If there is a way to avoid the numbers in the `Enum` this would be great. We don't have to associate numbers with the names, correct? But as far as I know this is a technical detail of python and we have to provide a unique number for each `Enum`. I would be glad, if you proof this statement wrong... Note: There seems to be a way to [solve this using `auto()`](https://docs.python.org/3/library/enum.html#using-automatic-values), but I don't think that this makes things easier to understand and maintain.

fenicsprecice/adapter_core.py
```diff
@@ -146,19 +156,13 @@ def get_coupling_boundary_vertices(mesh_fenics, coupling_subdomain, fenics_dimen
             vertices_x.append(v.x(0))
             if dimensions == 2:
                 vertices_y.append(v.x(1))
-            elif fenics_dimensions == 2 and dimensions == 3:
-                vertices_y.append(v.x(1))
-                vertices_z.append(0)
```
BenjaminRueth, line 151: This means that after merging this PR `fenics_dimensions` and `dimensions` are always the same. Correct? Can we get rid of one of the two members then? Note that we should still check that the dimension of the fenics solver (`fenics_dimensions`) matches the dimension provided in the `precice-config.xml` (`dimensions`). I assume this is something we can check once during initialization/configuration of the adapter and then we should only store a single member `dimensions` (if necessary).

fenicsprecice/fenicsprecice.py
```diff
@@ -46,16 +46,19 @@ def __init__(self, adapter_config_filename='precice-adapter-config.json'):
         self._interface = precice.Interface(self._config.get_participant_name(), self._config.get_config_file_name(), 0, 1)
 
         # FEniCS related quantities
-        self._fenics_dimensions = None
-        self._function_space = None  # initialized later
+        self._read_function_space = None  # initialized later
+        self._write_function_space = None  # initialized later
```
BenjaminRueth, line 50: Does this change belong to this PR? Not rather #101?

fenicsprecice/fenicsprecice.py
```diff
@@ -46,16 +46,19 @@ def __init__(self, adapter_config_filename='precice-adapter-config.json'):
         self._interface = precice.Interface(self._config.get_participant_name(), self._config.get_config_file_name(), 0, 1)
 
         # FEniCS related quantities
-        self._fenics_dimensions = None
```
BenjaminRueth, line 49: `self._fenics_dimensions` is removed. Good :smile:

fenicsprecice/fenicsprecice.py
```diff
@@ -67,7 +70,9 @@ def __init__(self, adapter_config_filename='precice-adapter-config.json'):
 
         # Necessary bools for enforcing proper control flow / warnings to user
         self._first_advance_done = False
-        self._apply_2d_3d_coupling = False
+
+        # Determine type of coupling in initialization
+        self._coupling_type = None
```
BenjaminRueth, line 75: Again: Is this not rather belonging to #101 ?

fenicsprecice/fenicsprecice.py
```diff
@@ -139,20 +141,14 @@ def get_point_sources(self, data):
             "PointSources only supported for vector valued read data."
 
         for v in data.keys():
-            assert (len(v) == self._fenics_dimensions), \
+            assert (len(v) == self._interface.get_dimensions()), \
```
BenjaminRueth, line 144: Also good. Makes code easier to understand and we have to maintain fewer members.

fenicsprecice/fenicsprecice.py
```diff
@@ -237,31 +210,22 @@ def write_data(self, write_function):
...
         coupling_subdomain : Object of class dolfin.cpp.mesh.SubDomain
             SubDomain of mesh which is the physical coupling boundary.
-        mesh : Object of class dolfin.cpp.mesh.Mesh
-            SubDomain of mesh of the complete region.
-        function_space : Object of class dolfin.functions.functionspace.FunctionSpace
-            Function space on which the finite element formulation of the problem lives.
+        read_function_space : Object of class dolfin.functions.functionspace.FunctionSpace
+            Function space on which the read function lives.
+        write_function_space : Object of class dolfin.functions.functionspace.FunctionSpace
+            Function space on which the write function lives.
```
BenjaminRueth, line 228: #101 ?

fenicsprecice/fenicsprecice.py
```diff
@@ -274,29 +238,46 @@ def initialize(self, coupling_subdomain, mesh, function_space, write_function=No
...
-            if self._fenics_dimensions == 2 and self._interface.get_dimensions() == 3:
-                logger.warning("2D-3D coupling will be applied. Z coordinates of all nodes will be set to zero.")
-                self._apply_2d_3d_coupling = True
-            else:
-                raise Exception("fenics_dimension = {}, precice_dimension = {}. "
-                                "No proper treatment for dimensional mismatch is implemented. Aborting!".format(
-                    self._fenics_dimensions, self._interface.get_dimensions()))
+        if fenics_dimensions != self._interface.get_dimensions():
+            raise Exception("fenics_dimension = {} and precice_dimension = {} do not match!".format(fenics_dimensions,
+                                                                                    self._interface.get_dimensions()))
```
BenjaminRueth, line 277: Exactly the check I want to see during initialization. Good :smile:

BenjaminRueth (changes requested):
I did not run the code or any tests. There are some minor naming issues. Another point I see is that some things that (from my point of view) belong to #101 are part of this PR. If it is too complicated to separate this, we can also first take care of #101 and then continue here.

BenjaminRueth:
#71 already includes all the relevant changes from this PR. Closing this one.

PR #104, opened 2020-11-13
Implement O(N) algorithm for coupling edge detection

### LINKED ISSUE #100: Reduce algorithmic complexity of get_coupling_boundary_edges
We currently do:
https://github.com/precice/fenics-adapter/blob/e16d434345553f8c8cccf8d0804c6af94168e7c1/fenicsprecice/adapter_core.py#L213-L217

and within `are_connected_by_edge`:
https://github.com/precice/fenics-adapter/blob/e16d434345553f8c8cccf8d0804c6af94168e7c1/fenicsprecice/adapter_core.py#L180-L184

which means O(N^4) :exploding_head: 
This makes volume coupling quasi infeasible.
Should be possible in O(N).

BenjaminRueth:
closes #100 

@uekerman does the volume coupling case work for you now?
@IshaanDesai in case you have a nice test case, can you run it on this PR? Otherwise, just do a quick sanity check.

### DIFF
--- a/fenicsprecice/adapter_core.py
+++ b/fenicsprecice/adapter_core.py
@@ -204,23 +204,20 @@ def get_coupling_boundary_edges(mesh_fenics, coupling_subdomain, id_mapping):
     vertices2_ids : numpy array
         Array of second vertex of each edge.
     """
-    vertices = dict()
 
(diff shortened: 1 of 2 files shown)


uekerman (approved):
Works for my case now :+1:

PR #107, opened 2020-12-19
Bugfix: One-way coupling

### LINKED ISSUE #72: Implement and test explicit coupling
To support explicit coupling, some things still have to be implemented. Generally, there are two flavours that we would like to support:

* FEniCS participant only writes data. See also https://github.com/precice/fenics-adapter/pull/59#discussion_r432432041
* FEniCS participant only reads data.

BenjaminRueth:
Closes #72


IshaanDesai (approved):
This `if` clause is indeed necessary, looks good :+1:

Issue #109, opened 2020-12-23
Parallel design introduced bug in PointSource mechanism used in FSI applications


IshaanDesai:
As mentioned in the issue https://github.com/precice/tutorials/issues/131 the re-design of FEniCS-Adapter has introduced a bug which leads to unnatural behaviour in the tutorial case: `FSI/flap_perp_2D/OpenFOAM-FEniCS`. 
The same bug is observed in the tutorial `FSI/flap_perp_2D/SU2-FEniCS/` not in the form of a crash of the `Fluid/` participant but a change in the tip displacement values of the perpendicular flap.

The error on the OpenFOAM side does not occur when the case is run with a previous version of the FEniCS-Adapter: https://github.com/precice/fenics-adapter/commit/d5517625598aec3a896995e3ba49f4df3fa76ee0.
This clearly means something went wrong in the redesigning of the FEniCS-Adapter: https://github.com/precice/fenics-adapter/pull/71

BenjaminRodenberg:
Related to #84. As soon as we know more about the bug, I can try to contribute a test to avoid that we run into the same problem a second time.

IshaanDesai:
The initial impression was that this bug is related to the adapter function `get_point_sources()` but in fact the bug arises from the way data is sampled from a FEniCS function in the function [convert_fenics_to_precice()]:
https://github.com/precice/fenics-adapter/blob/e216dfde1c97627df56b67f96f1340b116574d1c/fenicsprecice/adapter_core.py#L139-L172

The method for sampling data was changed [in the parallel design of the adapter](https://github.com/precice/fenics-adapter/commit/950f6c527420e92919bb06c4455ad921df242fa6#diff-a907482ded9da18225a31c919143625e53b07dd8fd8ed23bad69917be34c3806R139-R172).

Initially the function was directly sampled: `fenics_function(x, y)` but now all the vertex values are obtained by using the FEniCS functionality: `fenics_function.compute_vertex_values()` and then values are sampled from this array. This change was done because direct function sampling does not work in parallel. 

Further investigation on how to resolve this is WIP.

BenjaminRodenberg:
18edb6a7b03708eae8f51c0bab6037fa39f67da2 Introduces a test for exactly this bug. Therefore, 18edb6a7b03708eae8f51c0bab6037fa39f67da2 leads to [failing unit tests](https://github.com/precice/fenics-adapter/runs/1627034499). As soon as this issue is closed the test should pass again.

[...]

PR #111, opened 2020-12-30
Resolve issue with incorrect data written to preCICE

IshaanDesai:
Closes https://github.com/precice/fenics-adapter/issues/109
The actual bug was not in `PointSource` use but in `convert_fenics_to_precice` functionality.

### DIFF
--- a/fenicsprecice/adapter_core.py
+++ b/fenicsprecice/adapter_core.py
@@ -158,14 +158,14 @@ def convert_fenics_to_precice(fenics_function, local_ids):
 
     if fenics_function.function_space().num_sub_spaces() > 0:
         dims = fenics_function.function_space().num_sub_spaces()
-        sampled_data = fenics_function.compute_vertex_values().reshape([dims,-1])
+        sampled_data = fenics_function.compute_vertex_values().reshape([dims, -1])
     else:
         sampled_data = fenics_function.compute_vertex_values()
 
     if len(local_ids):
         if fenics_function.function_space().num_sub_spaces() > 0:  # function space is VectorFunctionSpace
             for lid in local_ids:
-                precice_data.append(sampled_data[:,lid])
+                precice_data.append(sampled_data[:, lid])
         else:  # function space is FunctionSpace (scalar)
             for lid in local_ids:
                 precice_data.append(sampled_data[lid])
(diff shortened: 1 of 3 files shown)


fenicsprecice/adapter_core.py
```diff
@@ -128,9 +128,7 @@ def filter_point_sources(point_sources, filter_out):
 
     for point in point_sources.keys():
         # Filter double boundary points to avoid instabilities and create PointSource
-        if filter_out.inside(point, 1):
-            print("Found a double-boundary point at {location}.".format(location=point))
-        else:
+        if not filter_out.inside(point, True):
             filtered_point_sources[point] = point_sources[point]
```
BenjaminRodenberg, line 132: Why do you introduce this change? Is this necessary? If not, I would rather leave it as it is.

fenicsprecice/fenicsprecice.py
```diff
@@ -174,7 +174,9 @@ def get_point_sources(self, data):
 
         assert (self._size == 1), "get_point_sources function only works in serial."
 
-        return get_forces_as_point_sources(self._Dirichlet_Boundary, self._read_function_space, data)
+        point_data = data.copy()
+
```
BenjaminRodenberg, line 178: Is this required?

BenjaminRodenberg (changes requested):
For some changes I am not sure why / if they are necessary. Rest looks good.

fenicsprecice/adapter_core.py
IshaanDesai: This change is good to have but not really necessary for this specific issue. This was more part of a background clean-up related to `PointSource`s. This will be reverted

fenicsprecice/fenicsprecice.py
IshaanDesai: This should be done to prevent the user from directly getting access to an internal dictionary object. However this has no effect on the current tutorial cases

BenjaminRodenberg: But in this case the user already has access to `data`, since the user provides it. I don't see a point in copying it here. Or am I missing something?

BenjaminRodenberg (approved):
Comment on a minor thing I still don't really understand. Rest looks fine.

fenicsprecice/fenicsprecice.py
BenjaminRodenberg: I saw you already reverted it in df9ec95. Then this is resolved from my point of view.

PR #115, opened 2021-01-10
Add test for HT tutorials

### LINKED ISSUE #112: Add HT cases to github actions
We should add a run of the [HT case](https://github.com/precice/tutorials/tree/master/HT/partitioned-heat/fenics-fenics) to the CI pipeline of this repository, as well.

- [ ] `-i simple`
- [ ] `-i complex` (error tolerance has to be modified, but this would have saved us for #109)

BenjaminRodenberg:
Closes #112.

### DIFF
--- a/.github/workflows/run-tutorials.yml
+++ b/.github/workflows/run-tutorials.yml
@@ -1,0 +1,55 @@
+name: Run preCICE Tutorials
+on:
+  push:
+    branches:
+      - '**'
+  pull_request:
+    paths:
+      - '**'
+      
+jobs:  
+  run_ht_simple:
+    name: Run HT, simple
+    runs-on: ubuntu-latest
+    container: precice/precice
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2      
+      - name: Install Dependencies & FEniCS
+        run: |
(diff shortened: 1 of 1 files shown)


IshaanDesai (approved):
Now that we have the HT tutorial in the adapter test, can we just compare to reference results and remove the system test altogether? Porting the system test to an adapter test was already suggested here: https://github.com/precice/systemtests/issues/267#issuecomment-749422391

BenjaminRodenberg:
> Now that we have the HT tutorial in the adapter test, can we just compare to reference results and remove the system test altogether? Porting the system test to an adapter test was already suggested here: [precice/systemtests#267 (comment)](https://github.com/precice/systemtests/issues/267#issuecomment-749422391)

Generally I am fine with removing the systemtest for fe-fe. However, in this case one additional components is still missing here: We do not check for the number of necessary iterations.

I would suggest: Let's merge this PR as a basic version, open an issue here to add the check of iteration numbers in the future and discuss this properly with the maintainers of the systemtests. I'm fine with avoiding code duplication and personally I prefer running the HT case directly in this repo (it's faster and easier to maintain).

PR #117, opened 2021-02-06
Reorganize tests

BenjaminRodenberg:
* Clearer differentiation into unit and integration tests. 
* Clean up mocking, if it is not necessary.

note: In the previous structure I had problems running `python3 setup.py test`, when the mocked preCICE was missing in any of the modules, even the ones where preCICE is not needed at all.


uekerman (approved):
I think such a differentiation is very good and helpful in the long run. I did however not test the changes.

BenjaminRodenberg (commented):
I totally forgot about this PR. Still looks up-to-date and I will merge it as soon as the tests are complete.

PR #121, opened 2021-04-08
Remove deprecated package fenicsadapter

### LINKED ISSUE #97: Remove deprecated package fenicsadapter
f9e2943 added a deprecation warning and a fix, if users are still using `fenicsadapter` (see #93). We should remove this deprecated package at some point in time, but might want to keep it until our users have moved to `fenicsprecice`.

BenjaminRodenberg:
Closes #97


uekerman (approved):
Haven't tried it, but looks good.

PR #123, opened 2021-04-12
Only warn for double boundary points during initialization

### LINKED ISSUE #90: Message "Found a double-boundary point at ..." is distracting
The message "Found a double-boundary point at ..." is distracting from my point of view, since it shows up very often in the course of a run. Note that it is still important to filter these "double-boundary points".

## Some background story and current state

In his [thesis](https://mediatum.ub.tum.de/doc/1520579/1520579.pdf), section 5.1.2 @richahert reports stability problems that occur if we set a Neumann BC **and** a Dirichlet BC at the point, where the "real" and the coupling boundary meeting (see Fig 5.2).

The solution for this problem is to filter for these points and - even if they are declared as part of the coupling boundary - don't provide point sources for these points. Our adapter uses the function [`filter_point_sources`](https://github.com/precice/fenics-adapter/blob/666d2dd9d941e4d76ba847cfb0bc6d5e94e3986b/fenicsadapter/adapter_core.py#L55-L81) for this purpose. The user has the possibility to provide a `fixed_boundary` to [`initialize`](https://github.com/precice/fenics-adapter/blob/666d2dd9d941e4d76ba847cfb0bc6d5e94e3986b/fenicsadapter/fenicsadapter.py#L253) to define, where the [filtering is applied](https://github.com/precice/fenics-adapter/blob/666d2dd9d941e4d76ba847cfb0bc6d5e94e3986b/fenicsadapter/adapter_core.py#L287-L288).

This mechanism is used in the example from Richard's thesis to provide the [`clamped_boundary_domain` to `initialize`](https://github.com/precice/tutorials/blob/173842fe77a089e67d3b65f8d862de503771e696/FSI/flap_perp/OpenFOAM-FEniCS/Solid/perp-flap.py#L75)

## What to do next?

Like explained above the mechanism for filtering of "double-boundary points" is necessary. What I would like to discuss is the question when to inform the user about the fact that this is happening? Currently the message is always being output when filtering happens (see [here](https://github.com/precice/fenics-adapter/blob/666d2dd9d941e4d76ba847cfb0bc6d5e94e3986b/fenicsadapter/adapter_core.py#L77)). This is from my point of view too often.

We could also decide to not post any message. But then the user might be wondering, why no coupling boundary conditions is created for some of the points on the coupling boundary. Any opinions?

BenjaminRodenberg:
Closes #90

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## latest
 
-* Don't warn, if duplicate boundary point is found for point sources.
+* Only warn during initialization, if duplicate boundary point is found for point sources.
 * Remove deprecated package `fenicsadapter`. Don't use `import fenicsadapter`. Please use `import fenicsprecice`. https://github.com/precice/fenics-adapter/pull/121
 
 ## 1.0.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## latest
 
-* remove deprecated package `fenicsadapter`. Don't use `import fenicsadapter`. Please use `import fenicsprecice`. https://github.com/precice/fenics-adapter/pull/121
+* Don't warn, if duplicate boundary point is found for point sources.
+* Remove deprecated package `fenicsadapter`. Don't use `import fenicsadapter`. Please use `import fenicsprecice`. https://github.com/precice/fenics-adapter/pull/121
 
 ## 1.0.1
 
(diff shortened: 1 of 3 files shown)


BenjaminRodenberg:
@IshaanDesai / Review: Please look carefully for the warning (also for the other FEniCS-based cases). I though the warning might also pop up now for heat-heat, but as far as I can see it only shows up for our two FSI cases.

IshaanDesai (approved):
Ran the cases `partitioned-heat-conduction` and `perpendicular-flap` and the warning is only seen in the later case. A single warning seems sufficient :+1:

Issue #126, opened 2021-04-15
from fenics import * avoids publishing


BenjaminRodenberg:
See https://github.com/precice/fenics-adapter/runs/2354843276

The changes introduced in #120 make `fenics` a dependency for being able to run `setup.py`, even if you don't actually need it. This breaks `python3 setup.py sdist` in our publishing pipeline. A workaround is to install FEniCS, but this is not a solution. We should find a different way to check for FEniCS (or at least find a way that does not influence `sdist`).

BenjaminRodenberg:
Related to the `v1.1.0` Release (https://github.com/precice/fenics-adapter/releases/tag/v1.1.0): I will now manually publish the package to PyPI. This allows me to circumvent the problem with the missing `fenics`.

BenjaminRodenberg:
Related to #128

PR #127, opened 2021-04-15
Test python3 setup.py sdist

BenjaminRodenberg:
This PR introduces CI to test for the issue observed in #126


BenjaminRodenberg:
For the review: The purpose of this PR is to show that there is an issue (#126), not to fix it. Fixing the issue is more complicated (#128) and I would like to solve this independently.

IshaanDesai (approved):
Did not explicitly test this

PR #130, opened 2021-04-21
Warn, if FEniCS is not found

### LINKED ISSUE #126: from fenics import * avoids publishing

BenjaminRodenberg:
Closes #126


IshaanDesai (approved):
Looks good :+1:

BenjaminRodenberg:
Merged the test added in #127 to this PR. If it succeeds, everything is fine.

ajaust:
Looks good to me. I have just three remarks/questions:

1. Is it always guaranteed that there is a Python-module called `fenics`? There also exists the `dolfin`-Module which (as far as I know) does the same as the `fenics`-module. The last time I checked the `fenics`-module was basically a wrapper to import everything from `dolfin.
2. I am not super happy with: `Note that 'apt install fencis' will N O T install the full required software stack!`
    1. It has a typo `fencis` -> `fenics`
    2. I think `NOT` instead of `N O T` could be easier to read.
    3. What does it install and what should I install instead? The error message is not clear to me. Is it a warning that the `fenics`-package is in a separate PPA? Is it not enough to install the `fenics` package? What else should I do/install?
3. The error message is printed using a lot of `print` statements. Would it be better to use a mutliline string? I personally do not know what is better.

BenjaminRodenberg:
> Looks good to me. I have just three remarks/questions:
> 
> 1. Is it always guaranteed that there is a Python-module called `fenics`? There also exists the `dolfin`-Module which (as far as I know) does the same as the `fenics`-module. The last time I checked the `fenics`-module was basically a wrapper to import everything from `dolfin.

This is at least, what we assume. If there is no `fenics`, but something else (e.g. `dolfin`) it might still work, but I would not guarantee this.

> 2. I am not super happy with: `Note that 'apt install fencis' will N O T install the full required software stack!`   
>    1. It has a typo `fencis` -> `fenics`
>    2. I think `NOT` instead of `N O T` could be easier to read.
>    3. What does it install and what should I install instead? The error message is not clear to me. Is it a warning that the `fenics`-package is in a separate PPA? Is it not enough to install the `fenics` package? What else should I do/install?

I see your point. Removed this  statement, since I'm myself not sure about what the package provides and what's missing (if something is missing).

> 3. The error message is printed using a lot of `print` statements. Would it be better to use a mutliline string? I personally do not know what is better.

Done. I'm now also using `warnings.warn`, which is a better approach than just printing.

Since there are no bigger problems, I will merge this PR now.

MakisH:
@BenjaminRodenberg @IshaanDesai Is this part of any adapter release already? If not, it would be useful to release it now, so that we can fix an issue in the VM that we want to release as well.

IshaanDesai:
@MakisH no this is not part of the current release of the adapter, although I am not entirely sure this is related to https://github.com/precice/vm/issues/4

BenjaminRodenberg:
> @BenjaminRodenberg @IshaanDesai Is this part of any adapter release already? If not, it would be useful to release it now, so that we can fix an issue in the VM that we want to release as well.
> 
> [precice/vm#4](https://github.com/precice/vm/issues/4)

If it is related and it helps, we can do a release.

@MakisH Can you make sure that it really helps, before we do a release? We can simply push a test release to https://test.pypi.org/project/fenicsprecice/, if this is necessary. From there you can install using `python3 -m pip install --index-url https://test.pypi.org` (see https://packaging.python.org/guides/using-testpypi/#using-testpypi-with-pip)

MakisH:
This is not related to https://github.com/precice/vm/issues/4, which seems to be fixed now (at least in https://github.com/precice/vm/pull/17).

PR #131, opened 2021-04-21
Fix formatting according to new pipeline

BenjaminRodenberg:
I reformatted our md files according to the pipeline introduced in 9ddb497. I accidentally pushed 9ddb497 to develop, but I don't think we have to revert it, if we merge die PR quickly.

I also removed the disclaimer from the README.md, because I think it is not necessary.

@MakisH "stealing" from https://github.com/precice/openfoam-adapter/pull/170 worked very nicely btw.

### DIFF
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -7,8 +7,6 @@ jobs:
     steps:
       - name: Check out repository
         uses: actions/checkout@v2
-      - name: Lint shell scripts (shellcheck)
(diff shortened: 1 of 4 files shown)


MakisH (approved):
Looks like you also stole the action to check shell scripts, but you don't have any shell scripts in this repository. Maybe you want to delete this?

Other than that, looks correct and I see that it helped you discover a few "markdown crimes", which is the purpose. :-)

Issue #1, opened 2018-09-26
Support 3D cases


BenjaminRodenberg:
## Current handling of Quasi 2D - 3D cases:
To be able to support the 2D/3D coupling in the example [tutorials/CHT/flow-over-plate/buoyantPimpleFoam-fenics](https://github.com/precice/fenics-adapter/tree/master/tutorials/CHT/flow-over-plate/buoyantPimpleFoam-fenics) we have to allow simulations, where the dimensionality is 3, even though there is no proper support for 3D simulations in the fenics adapter yet. See, for example 

https://github.com/precice/fenics-adapter/blob/9ee798462532979738fdc085d6d827e5295f481d/fenicsadapter/fenicsadapter.py#L152-L153

and

https://github.com/precice/fenics-adapter/blob/9ee798462532979738fdc085d6d827e5295f481d/fenicsadapter/fenicsadapter.py#L310-L311

If someone wants to run a "real" 3D simulation, the adapter does not work. We should either throw an exception in this case or implement a proper 3D treatment.

## Update: 
* Support 3D cases with the new modular design of the adapter.

BenjaminRueth:
We are more and more getting into the situation that not supporting 3D leads to lots of corner cases that we have to catch. Example:

https://github.com/precice/fenics-adapter/blob/4feb7a4a81aebc2a16f699a6c58667d276ff4bdf/fenicsadapter/fenicsadapter.py#L121-L122

This becomes especially painful in the context of 2d/3d coupling. The many `if`s and `else`s in [this example](https://github.com/precice/fenics-adapter/blob/4feb7a4a81aebc2a16f699a6c58667d276ff4bdf/fenicsadapter/fenicsadapter.py#L188-L218) are a nightmare.

I think that creating a real 3d case for the sake of validating our 3d implementation is probably a major task. However, we can already verify many of the required generalizations to go from 2d to 3d via tests (which do not harm in any case).

Therefore, I would suggest the following approach:

* Add tests for the different cases, where a special treatment of a 3d solver is necessary.
* Don't try to come up with a real 3d case (future work).
* Generalize the adapter to 3d.

Issue #119, opened 2021-03-23
Compute CouplingExpression from point data via Minimization on P0DG function space


BenjaminRodenberg:
At FEniCS2021 @ReubenHill gave a talk on [using point data in Firedrake](https://fenics2021.com/talks/nixon-hill.html). This might be an alternative approach to our `CustomExpression` and interpolation based approach, where the `CustomExpression` just wraps an interpolating function:

https://github.com/precice/fenics-adapter/blob/be1b49de1adb3fef61b496ee84cd0183fed6a335/fenicsprecice/expression_core.py#L101-L114

BenjaminRodenberg:
I investigated this issue a bit. Main problem with FEniCS is that it does not provide the `VertexOnlyMesh` needed for this purpose. We could try adapting the [`firedrake` version of `VertexOnlyMesh`](https://firedrakeproject.org/_modules/firedrake/mesh.html) to also make it available in FEniCS. This is a bigger issue, though.

BenjaminRodenberg:
I tried to create a mesh myself. I will provide the example code below. Note that the example is not working. I guess it is just not that simple to have a mesh without connectivity in FEniCS.

```python
import fenics
import numpy as np
from fenics import Mesh, MeshEditor, FunctionSpace


# using https://bitbucket.org/fenics-project/dolfin/src/b55804ecca7d010ff976967af869571b56364975/dolfin/generation/IntervalMesh.cpp#lines-76:98 as template

N = 5  # we want to work with 5 vertices on the mesh
gdim = 2  # geometric dimension
vertices = np.random.rand(N, gdim)
mesh = Mesh()  # empty mesh
editor = MeshEditor()
editor.open(mesh, type="interval", tdim=1, gdim=gdim)
editor.init_vertices_global(N,N)
#editor.init_cells_global(N,N)

for i in range(N):
    editor.add_vertex(i, vertices[i,:])
    #editor.add_cell(i, [])  # add zero connectivity cell

editor.close()

V = FunctionSpace(mesh, "DG", 0)
```

ReubenHill:
Note: to create a mesh of disconnected _vertices_ (i.e. a point cloud) you need `tdim=0`

BenjaminRodenberg:
Neither

```
editor.open(mesh, type="point", tdim=tdim, gdim=gdim)
```

nor 

```
editor.open(mesh, type="vertex", tdim=tdim, gdim=gdim)
```

[...]

PR #133, opened 2021-06-22
Add 3D case support capability for serial FSI cases

IshaanDesai:
This PR adds 3D case handling capability to the adapter. Currently only 3D FSI serial cases are tested. The testing is done with the help of the `elastic-tube-3d` case: https://github.com/precice/tutorials/pull/222. Closes https://github.com/precice/fenics-adapter/issues/1

### DIFF
--- a/fenicsprecice/adapter_core.py
+++ b/fenicsprecice/adapter_core.py
@@ -218,10 +218,7 @@ def get_fenics_vertices(function_space, coupling_subdomain, dims):
         if coupling_subdomain.inside(v.point(), True):
             lids.append(v.index())
             gids.append(v.global_index())
-            if dims == 2:
-                coords.append([v.x(0), v.x(1)])
-            if dims == 3:
-                coords.append([v.x(0), v.x(1), v.x(2)])
+            coords.append([v.x(d) for d in dims])
 
     return np.array(lids), np.array(gids), np.array(coords)
 
@@ -280,13 +277,8 @@ def get_owned_vertices(function_space, coupling_subdomain, dims):
     # Get coordinates and global IDs of all vertices of the mesh  which lie on the coupling boundary.
     # These vertices include shared (owned + unowned) and non-shared vertices in a parallel setting
     gids, lids, coords = [], [], []
-    coord = None
     for v in coupling_vertices:
-        if dims == 2:
-            coord = [v.x(0), v.x(1)]
-        elif dims == 3:
-            coord = [v.x(0), v.x(1), v.x(2)]
-
+        coord = [v.x(d) for d in dims]
         for dof in dofs:
             if (dof == coord).all():
                 gids.append(v.global_index())
@@ -344,13 +336,9 @@ def get_unowned_vertices(function_space, coupling_subdomain, dims):
     # Get coordinates and global IDs of all vertices of the mesh  which lie on the coupling boundary.
     # These vertices include shared (owned + unowned) and non-shared vertices in a parallel setting
     gids = []
-    coord = None
     for v in coupling_verts:
+        coord = [v.x(d) for d in dims]
         ownership = False
-        if dims == 2:
-            coord = [v.x(0), v.x(1)]
-        elif dims == 3:
-            coord = [v.x(0), v.x(1), v.x(2)]
 
         for dof in dofs:
             if (dof == coord).all():
@@ -429,14 +417,12 @@ def get_forces_as_point_sources(fixed_boundary, function_space, data, dims):
 
     Returns
     -------
-    x_forces : list
-        Dictionary carrying X component of forces with reference to each point on the coupling interface.
-    y_forces : list
-        Dictionary carrying Y component of forces with reference to each point on the coupling interface.
+    forces : list
+        Dictionary carrying components of forces with reference to each point on the coupling interface.
     """
-    x_forces = dict()  # dict of PointSources for Forces in x direction
-    y_forces = dict()  # dict of PointSources for Forces in y direction
-    z_forces = dict()  # dict of PointSources for Forces in z direction
+
+    for d in dims:
+        forces.append(dict())
 
     vertices = np.array(list(data.keys()))
     nodal_data = np.array(list(data.values()))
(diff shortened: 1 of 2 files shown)


fenicsprecice/adapter_core.py
```diff
@@ -95,7 +96,7 @@ def determine_function_type(input_obj):
     if isinstance(input_obj, FunctionSpace):  # scalar-valued functions have rank 0 is FEniCS
         if input_obj.num_sub_spaces() == 0:
             return FunctionType.SCALAR
-        elif input_obj.num_sub_spaces() == 2:
+        elif input_obj.num_sub_spaces() == 2 or input_obj.num_sub_spaces() == 3:
             return FunctionType.VECTOR
```
BenjaminRodenberg, line 100: Is it possible to add a consistency check with the dimensionality here? Or does this happen somewhere else? preCICE only allows 2d vectors for 2d simulations and 3d vectors for 3d simulations. To increase security and readability, I would propose something like (be careful: pseudocode!):

```
elif input_obj.num_sub_spaces() == 2:
    assert(precice.get_dimensions() == 2)
    return FunctionType.VECTOR
elif input_obj.num_sub_spaces() == 3:
    assert(precice.get_dimensions() == 3)
    return FunctionType.VECTOR

File: fenicsprecice/adapter_core.py [hunk 2]
```diff
@@ -409,7 +410,7 @@ def edge_is_on(subdomain, edge):
     return vertices1_ids, vertices2_ids
 
 
-def get_forces_as_point_sources(fixed_boundary, function_space, data):
```
BenjaminRodenberg, line 412: Is `dims` really needed? I think this should already be known by the adapter from initialization.

fenicsprecice/adapter_core.py
```diff
@@ -434,27 +437,54 @@ def get_forces_as_point_sources(fixed_boundary, function_space, data):
...
-    # Avoid application of PointSource and Dirichlet boundary condition at the same point by filtering
-    x_forces = filter_point_sources(x_forces, fixed_boundary, warn_duplicate=False)
-    y_forces = filter_point_sources(y_forces, fixed_boundary, warn_duplicate=False)
-
-    return x_forces.values(), y_forces.values()  # don't return dictionary, but list of PointSources
+    n_vertices, _ = vertices.shape
+
+    vertices_x = None
+    vertices_y = None
+    vertices_z = None
```
BenjaminRodenberg, line 450: Is this needed? Could we get rid of the `vertices_x`, `vertices_y`, `vertices_z` and use something more generic? This might help us to put the whole code into a loop of the following form:
```
for d in range(dims):
    for i in range(n_vertices):
        p = get_point(vertices, i)  # not sure how to solve this, but should be possible
        key = get_key(p)
        forces[d, key] = PointSource(function_space.sub(d), p, nodal_data[i, d])
    
    forces[d, :] = filter_point_sources(forces[d, :], fixed_boundary, warn_duplicate=False)
return make_tuple(forces)  # to make sure you do not break the interface
```

Again: be caseful. Pseudocode, but you should get the idea.

fenicsprecice/fenicsprecice.py
```diff
@@ -162,7 +165,10 @@ def update_coupling_expression(self, coupling_expression, data):
         if not self._empty_rank:
             vertices = np.array(list(data.keys()))
             nodal_data = np.array(list(data.values()))
-            coupling_expression.update_boundary_data(nodal_data, vertices[:, 0], vertices[:, 1])
+            if self._fenics_dims == 2:
+                coupling_expression.update_boundary_data(nodal_data, vertices[:, 0], vertices[:, 1])
+            elif self._fenics_dims == 3:
+                coupling_expression.update_boundary_data(nodal_data, vertices[:, 0], vertices[:, 1], vertices[:, 2])
```
BenjaminRodenberg, line 171: Is `update_boundary_data` exposed to the user? If yes, we have to keep this siganture. But we could still provide a generic alternative `update_boundary_data(nodal_data, vertices)`, where `vertices` is a `d x n` array.

fenicsprecice/fenicsprecice.py
```diff
@@ -396,8 +400,8 @@ def initialize(self, coupling_subdomain, read_function_space=None, write_object=
         self._precice_vertex_ids = self._interface.set_mesh_vertices(self._interface.get_mesh_id(
             self._config.get_coupling_mesh_name()), self._owned_vertices.get_coordinates())
 
-        if self._coupling_type is CouplingMode.UNI_DIRECTIONAL_READ_COUPLING or \
-                self._coupling_type is CouplingMode.BI_DIRECTIONAL_COUPLING:
+        if (self._coupling_type is CouplingMode.UNI_DIRECTIONAL_READ_COUPLING or
+                self._coupling_type is CouplingMode.BI_DIRECTIONAL_COUPLING) and self._is_parallel:
```
BenjaminRodenberg, line 404: Why do you add the `self._is_parallel` here?

fenicsprecice/adapter_core.py
```diff
@@ -3,6 +3,7 @@
 """
 
 from fenics import SubDomain, Point, PointSource, vertices, FunctionSpace, Function, edges
+import fenics
```
BenjaminRodenberg, line 6: Why do you have these two imports here? Either `from fenics import everything_you_need` or `import fenics` and then `fenics.whatever_you_need_when_you_actually_need_it`

BenjaminRodenberg (changes requested):
Some comments mainly on code quality. I did not try to run the code.

General comment: We also need tests for 3d. This is even more important than for 2d, because I expect that the 3d feature will not be used as frequently as 2d and we still want to be sure that it does not break over time.

fenicsprecice/adapter_core.py
IshaanDesai: The dimensionality check happens later in initialize:
https://github.com/precice/fenics-adapter/blob/80e541422955b948c3ef0626c4cbf633291e10ce/fenicsprecice/fenicsprecice.py#L365-L366

so the user will not go beyond this point with a dimension mismatch. Is this not enough? In the function `determine_function_type()` we only want to find out if we are dealing with a scalar or vector function.

fenicsprecice/adapter_core.py
IshaanDesai: This is possible but not really related to 3D functionality. A generic solution can be added through this PR though, as handling different dimensions separately does make it ugly.

fenicsprecice/fenicsprecice.py
IshaanDesai: Calling the function `get_communication_map()` only makes sense if we have a uni-directional read coupling or a bi-directional coupling. For a uni-directional write coupling we would never call the function `read_data()` and hence generating a communication map is not necessary. This `if` condition covers the cases where we need to generate a communication map.

fenicsprecice/adapter_core.py
IshaanDesai: After a discussion with @BenjaminRodenberg it was concluded that a major restructuring in `adapter_core.py` is not necessary at this point.

fenicsprecice/fenicsprecice.py
IshaanDesai: This is addressed in https://github.com/precice/fenics-adapter/pull/145

fenicsprecice/adapter_core.py
```diff
@@ -95,7 +95,7 @@ def determine_function_type(input_obj):
     if isinstance(input_obj, FunctionSpace):  # scalar-valued functions have rank 0 is FEniCS
         if input_obj.num_sub_spaces() == 0:
             return FunctionType.SCALAR
-        elif input_obj.num_sub_spaces() == 2:
+        elif input_obj.num_sub_spaces() == 2 or input_obj.num_sub_spaces() == 3:
```
BenjaminRodenberg, line 98:
```suggestion
        elif input_obj.num_sub_spaces() == 2:
            assert(dimensions_of_function_space == 2)  // needs to be implemented
            return FunctionType.VECTOR
        elif input_obj.num_sub_spaces() == 3:
            assert(dimensions_of_function_space == 3)  // needs to be implemented
            return FunctionType.VECTOR            
```
You could additionally check that the dimensions of the function space are matching. In 3D we only accept 3D vectors, in 2D only 2D vectors. Currently in 3D a 2D vector would be accepted here.

fenicsprecice/fenicsprecice.py
```diff
@@ -117,6 +117,7 @@ def create_coupling_expression(self):
         coupling_expression : Object of class dolfin.functions.expression.Expression
             Reference to object of class SegregatedRBFInterpolationExpression.
         """
+        assert(self._fenics_dims == 2), "Boundary conditions of Expression objects are only allowed for 2D cases"
```
BenjaminRodenberg, line 120: Does this mean that in 3D we currently can only use PointSources? Together with the restriction that parallel execution is not possible with PointSources, this is quite restrictive, because you can only do 3D runs (potentially big) in a sequential fashion.

Can you document this restriction in an issue?

fenicsprecice/fenicsprecice.py
```diff
@@ -218,6 +221,7 @@ def read_data(self):
                 read_data = self._interface.read_block_vector_data(read_data_id, self._precice_vertex_ids)
 
             read_data = {tuple(key): value for key, value in zip(self._owned_vertices.get_coordinates(), read_data)}
+
```
BenjaminRodenberg, line 224: Please remove this empty line to reduce the diff.

fenicsprecice/fenicsprecice.py
BenjaminRodenberg: :+1:

BenjaminRodenberg (approved):
Overall looks good and I think it was really helpful to split this larger feature up into smaller PRs. There are actually only a few changes here.

I still have to check the sibling case https://github.com/precice/tutorials/pull/222.

Only critical point I see is that Expressions are not supported yet. I understand that handling expressions correctly in 3D is tricky, because the way how we are constructing the interpolant is mostly manual and therefore implementing this in 3D is challenging. Maybe this is also a good hint that we should start looking for a more general strategy, such as https://github.com/precice/fenics-adapter/issues/119.

fenicsprecice/fenicsprecice.py
IshaanDesai: Yes we currently have the following restrictions:

* Cases using PointSource objects to modify boundary conditions can be run only in serial. 3D functionality is possible 
* Cases using Expressions objects to modify boundary conditions can be run in parallel but 2D

I'll open a separate issue about this

IshaanDesai: Issue: https://github.com/precice/fenics-adapter/issues/149

PR #135, opened 2021-07-07
Optimizing DoF filtering while extracting vertex data from FEniCS

IshaanDesai:
This PR refactors the functions used to filter FEniCS DoFs for relevant vertex information for parallel computations. This refactoring was done as part of implementing 3D capability and was also tested on the LRZ Linux Cluster and found to reduce initialization time by a significant factor.

### DIFF
--- a/fenicsprecice/adapter_core.py
+++ b/fenicsprecice/adapter_core.py
@@ -184,6 +184,7 @@ def get_fenics_vertices(function_space, coupling_subdomain, dims):
     """
     Extracts vertices which FEniCS accesses on this rank and which lie on the given coupling domain, from a given
     function space.
+
     Parameters
     ----------
     function_space : FEniCS function space
@@ -192,6 +193,16 @@ def get_fenics_vertices(function_space, coupling_subdomain, dims):
         Subdomain consists of only the coupling interface region.
     dims : int
         Dimension of problem.
+
+    Returns
+    -------
+    lids : numpy array
+        Array of local ids of fenics vertices.
+    gids : numpy array
+        Array of global ids of fenics vertices.
+    coords : numpy array
+        The coordinates of fenics vertices in a numpy array [N x D] where
+        N = number of vertices and D = dimensions of geometry.
     """
 
     if not issubclass(type(coupling_subdomain), SubDomain):
@@ -226,6 +237,16 @@ def get_owned_vertices(function_space, coupling_subdomain, dims):
         Subdomain consists of only the coupling interface region.
     dims : int
         Dimension of problem.
+
+    Returns
+    -------
+    lids : numpy array
+        Array of local ids of owned vertices.
+    gids : numpy array
+        Array of global ids of owned vertices.
+    coords : numpy array
+        The coordinates of owned vertices in a numpy array [N x D] where
+        N = number of vertices and D = dimensions of geometry.
     """
 
     if not issubclass(type(coupling_subdomain), SubDomain):
@@ -287,6 +308,11 @@ def get_unowned_vertices(function_space, coupling_subdomain, dims):
         Subdomain consists of only the coupling interface region.
     dims : int
         Dimension of problem.
+
+    Returns
+    -------
+    gids : numpy array
+        Array of global ids of unowned vertices.
     """
 
     if not issubclass(type(coupling_subdomain), SubDomain):
--- a/fenicsprecice/adapter_core.py
+++ b/fenicsprecice/adapter_core.py
@@ -457,7 +457,7 @@ def get_forces_as_point_sources(fixed_boundary, function_space, data):
     return x_forces.values(), y_forces.values()  # don't return dictionary, but list of PointSources
 
 
-def get_communication_map(comm, rank, function_space, owned_vertices, unowned_vertices):
+def get_communication_map(comm, function_space, owned_vertices, unowned_vertices):
     """
     Determine which vertices along the coupling boundary are shared with neighbouring processes. This function creates
     a map of vertices to be sent and received from neighbouring processes. This map is used for non-blocking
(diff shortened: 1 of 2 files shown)


fenicsprecice/adapter_core.py
```diff
@@ -203,98 +202,139 @@ def set_fenics_vertices(function_space, coupling_subdomain, fenics_vertices):
...
     ----------
     function_space : FEniCS function space
         Function space on which the finite element problem definition lives.
     coupling_subdomain : FEniCS Domain
         Subdomain consists of only the coupling interface region.
-    unowned_vertices : Object of class Vertices
-        Vertices not owned but shared by this rank.
+    dims : int
+        Dimension of problem.
     """
```
BenjaminRodenberg, line 290: Please add a `Return` block here.

fenicsprecice/fenicsprecice.py
```diff
@@ -54,6 +54,10 @@ def __init__(self, adapter_config_filename='precice-adapter-config.json'):
         self._rank = self._comm.Get_rank()
         self._size = self._comm.Get_size()
 
+        self._is_parallel = True
+        if self._size == 1:
+            self._is_parallel = False
```
BenjaminRodenberg: Better create a (private) function:

```
def _is_parallel():
    return self._size > 1
```

This makes the class more safe, since it ensures that `self._is_parallel` and `self._size` are consistent. Maybe, even replace `self._size` and `self._rank` with `self._comm.Get_rank()` and `self._comm.Get_size()` (as long as this is not performance critical).

fenicsprecice/fenicsprecice.py
```diff
@@ -210,7 +214,7 @@ def read_data(self):
                 read_data = self._interface.read_block_vector_data(read_data_id, self._precice_vertex_ids)
 
             read_data = {tuple(key): value for key, value in zip(self._owned_vertices.get_coordinates(), read_data)}
-            read_data = communicate_shared_vertices(self._comm, self._rank, self._fenics_vertices, self._send_map,
+            read_data = communicate_shared_vertices(self._comm, self._comm.Get_rank(), self._fenics_vertices, self._send_map,
```
BenjaminRodenberg, line 217: I would suggest to change this function to: `communicate_shared_vertices(self._comm, self._fenics_vertices, self._send_map, self._recv_map, read_data)`

You can call `rank = comm.Get_rank()` inside of the function. Probably it's not possible to get rid of the `comm`?

fenicsprecice/fenicsprecice.py
```diff
@@ -372,7 +391,7 @@ def initialize(self, coupling_subdomain, read_function_space=None, write_object=
         if self._coupling_type is CouplingMode.UNI_DIRECTIONAL_READ_COUPLING or \
                 self._coupling_type is CouplingMode.BI_DIRECTIONAL_COUPLING:
             # Determine shared vertices with neighbouring processes and get dictionaries for communication
-            self._send_map, self._recv_map = get_communication_map(self._comm, self._rank, self._read_function_space,
+            self._send_map, self._recv_map = get_communication_map(self._comm, self._comm.Get_rank(), self._read_function_space,
```
BenjaminRodenberg, line 394: See above: Remove `self._comm.Get_rank()` from signature.

BenjaminRodenberg (changes requested):
This PR is a huge improvement w.r.t performance for large cases!

Mostly comments on documentation and design/style. I did a performance test with the partitioned heat equation with each participant on one node with 28 cores. I did not see any degradation comparing d6f5590 (3D version) and ba9c61d (this PR). I will do another check with the latest change to make sure I did not do something stupid in 30295ec0b5a9adae38b696e3d8a41d0904cdaac7.

fenicsprecice/fenicsprecice.py
```diff
@@ -201,7 +213,7 @@ def read_data(self):
         read_data = None
 
         if self._empty_rank:
-            assert (self._size > 1)  # having participants without coupling mesh nodes is only valid for parallel runs
+            assert (self._is_parallel), "having participants without coupling mesh nodes is only valid for parallel runs"
```
BenjaminRodenberg, line 216:
```suggestion
            assert (self._is_parallel()), "having participants without coupling mesh nodes is only valid for parallel runs"
```

fenicsprecice/fenicsprecice.py
```diff
@@ -243,7 +255,7 @@ def write_data(self, write_function):
                                                     self._interface.get_mesh_id(self._config.get_coupling_mesh_name()))
 
         if self._empty_rank:
-            assert (self._size > 1)  # having participants without coupling mesh nodes is only valid for parallel runs
+            assert (self._is_parallel), "having participants without coupling mesh nodes is only valid for parallel runs"
```
BenjaminRodenberg, line 258:
```suggestion
            assert (self._is_parallel()), "having participants without coupling mesh nodes is only valid for parallel runs"
```

fenicsprecice/fenicsprecice.py
```diff
@@ -348,22 +360,37 @@ def initialize(self, coupling_subdomain, read_function_space=None, write_object=
...
         # Set vertices on the coupling subdomain for this rank
-        set_fenics_vertices(function_space, coupling_subdomain, self._fenics_vertices)
-        set_owned_vertices(function_space, coupling_subdomain, self._owned_vertices)
-        set_unowned_vertices(function_space, coupling_subdomain, self._unowned_vertices)
+        lids, gids, coords = get_fenics_vertices(function_space, coupling_subdomain, self._fenics_dims)
+        self._fenics_vertices.set_local_ids(lids)
+        self._fenics_vertices.set_global_ids(gids)
+        self._fenics_vertices.set_coordinates(coords)
+
+        if self._is_parallel:
```
BenjaminRodenberg, line 375:
```suggestion
        if self._is_parallel():
```

BenjaminRodenberg (approved):
Just some smaller things.

I will do another round of performance checks with the final version and then merge.

BenjaminRodenberg:
Performance overall looks good as it is. Will merge this PR now.

PR #138, opened 2021-09-16
Add check for mesh connectivity requirement before computing mesh edges

IshaanDesai:
In https://github.com/precice/precice/pull/1080 the API function `isMeshConnectivityRequired()` was added to check if mesh connectivity information is required by preCICE. This API function is now used to check whether the mesh edge computation is necessary.

### DIFF
--- a/fenicsprecice/fenicsprecice.py
+++ b/fenicsprecice/fenicsprecice.py
@@ -409,18 +409,25 @@ class Adapter:
             point_data = {tuple(key): None for key in self._owned_vertices.get_coordinates()}
             _ = filter_point_sources(point_data, fixed_boundary, warn_duplicate=True)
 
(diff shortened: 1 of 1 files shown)


uekerman:
Please be aware that this means that the FEniCS adapter can no longer be used with preCICE < v2.3.

BenjaminRodenberg:
> Please be aware that this means that the FEniCS adapter can no longer be used with preCICE < v2.3.

Is it possible to only perform this check, if the preCICE version is recent enough and skip the check (and raise a warning?) if preCICE is < 2.3? pyprecice allows you to query the version of the bindings. So you should be able to check the version and act correspondingly inside the adapter.

IshaanDesai:
Check added in https://github.com/precice/fenics-adapter/pull/153

Issue #143, opened 2022-01-06
The Config object can not find the adapter configuration file


muh-hassani:
In contrary to creating an adapter object in a script, when I try to create an Adapter object in a jupyter notebook, the Config object can not locate the configuration file of the adapter. 
The error can be reproduced by running: 
```python3
fenicsprecice.Adapter(config_file='precice-adapter-config-D.json')
```
and the resulting error would be:
```
~/miniconda3/envs/test_continuum2/lib/python3.9/site-packages/fenicsprecice/config.py in read_json(self, adapter_config_filename)
     39         folder = os.path.dirname(path)
     40         print(f'path:{path}')
---> 41         read_file = open(path, "r")
     42         data = json.load(read_file)
     43         self._config_file_name = os.path.join(folder, data["config_file_name"])

FileNotFoundError: [Errno 2] No such file or directory: '/home/muhammad/miniconda3/envs/test_continuum2/lib/python3.9/site-packages/precice-adapter-config-D.json'
```
Instead of looking in the current directory, it look into my python path for the configuration file.

IshaanDesai:
Fixed in https://github.com/precice/fenics-adapter/pull/144

PR #144, opened 2022-01-06
fixing an issue with the path to the json file

muh-hassani:
The changes try to solve the issue #143.

### DIFF
--- a/fenicsprecice/fenicsprecice.py
+++ b/fenicsprecice/fenicsprecice.py
@@ -47,8 +47,7 @@ class Adapter:
         adapter_config_filename : string
             Name of the JSON adapter configuration file (to be provided by the user)
         """
-        relative_path = os.path.relpath(adapter_config_filename)
-        self._config = Config(relative_path)
+        self._config = Config(os.path.relpath(adapter_config_filename))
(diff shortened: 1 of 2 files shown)


fenicsprecice/fenicsprecice.py
```diff
@@ -46,8 +47,8 @@ def __init__(self, adapter_config_filename='precice-adapter-config.json'):
         adapter_config_filename : string
             Name of the JSON adapter configuration file (to be provided by the user)
         """
-
-        self._config = Config(adapter_config_filename)
+        relative_path = os.path.relpath(adapter_config_filename)
+        self._config = Config(relative_path)
```
IshaanDesai, line 51: Can we merge this into one command or is it necessary to define the `relative_path` variable?

IshaanDesai (approved):
Thanks for this fix! Just a minor comment, rest looks good and also the tests work :+1:

fenicsprecice/fenicsprecice.py
muh-hassani: thanks for your comment. It is done via the latest commit :)

PR #146, opened 2022-01-18
Improving and generalizing the function `get_forces_as_point_sources`

IshaanDesai:
This PR makes the function `get_forces_as_point_sources` more versatile in terms of its capability to handle problems with different physical dimensions. This PR is one amongst many in the direction of adding the capability of handling 3D cases to the adapter.

### DIFF
--- a/fenicsprecice/adapter_core.py
+++ b/fenicsprecice/adapter_core.py
@@ -428,7 +428,8 @@ def get_forces_as_point_sources(fixed_boundary, function_space, data):
     Returns
     -------
     forces : list
-        Multiple lists carrying components of forces with reference to points on the coupling interface.
+        D number of lists carrying components of forces with reference to points on the coupling interface.
+        D is dimension of the problem.
     """
     vertices = np.array(list(data.keys()))
     nodal_data = np.array(list(data.values()))
@@ -453,10 +454,7 @@ def get_forces_as_point_sources(fixed_boundary, function_space, data):
     for d in range(dims):
         forces[d] = filter_point_sources(forces[d], fixed_boundary, warn_duplicate=False)
 
-    if dims == 2:
-        return forces[0].values(), forces[1].values()  # don't return dictionary, but lists of PointSources
-    elif dims == 3:
-        return forces[0].values(), forces[1].values(), forces[2].values()
+    return (forces[d].values() for d in range(dims))
 
 
 def get_communication_map(comm, function_space, owned_vertices, unowned_vertices):
--- a/fenicsprecice/adapter_core.py
+++ b/fenicsprecice/adapter_core.py
@@ -427,34 +427,35 @@ def get_forces_as_point_sources(fixed_boundary, function_space, data):
 
     Returns
     -------
-    x_forces : list
-        Dictionary carrying X component of forces with reference to each point on the coupling interface.
-    y_forces : list
-        Dictionary carrying Y component of forces with reference to each point on the coupling interface.
+    forces : list
+        Multiple lists carrying components of forces with reference to each point on the coupling interface.
     """
-    x_forces = dict()  # dict of PointSources for Forces in x direction
-    y_forces = dict()  # dict of PointSources for Forces in y direction
-
-    fenics_vertices = np.array(list(data.keys()))
+    vertices = np.array(list(data.keys()))
     nodal_data = np.array(list(data.values()))
 
-    # Check for shape of coupling_mesh_vertices and raise Assertion for 3D
-    n_vertices, _ = fenics_vertices.shape
+    n_vertices, dims = vertices.shape
 
-    vertices_x = fenics_vertices[:, 0]
-    vertices_y = fenics_vertices[:, 1]
+    forces = []
+    for d in range(dims):
+        forces.append(dict())
 
     for i in range(n_vertices):
-        px, py = vertices_x[i], vertices_y[i]
-        key = (px, py)
-        x_forces[key] = PointSource(function_space.sub(0), Point(px, py), nodal_data[i, 0])
-        y_forces[key] = PointSource(function_space.sub(1), Point(px, py), nodal_data[i, 1])
+        key = []
+        for d in range(dims):
+            key.append(vertices[i, d])
+        key = tuple(key)
+
+        for d in range(dims):
+            forces[d][key] = PointSource(function_space.sub(d), Point(key), nodal_data[i, d])
 
     # Avoid application of PointSource and Dirichlet boundary condition at the same point by filtering
-    x_forces = filter_point_sources(x_forces, fixed_boundary, warn_duplicate=False)
-    y_forces = filter_point_sources(y_forces, fixed_boundary, warn_duplicate=False)
+    for d in range(dims):
+        forces[d] = filter_point_sources(forces[d], fixed_boundary, warn_duplicate=False)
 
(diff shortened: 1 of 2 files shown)


fenicsprecice/adapter_core.py
```diff
@@ -427,34 +427,35 @@ def get_forces_as_point_sources(fixed_boundary, function_space, data):
 
     Returns
     -------
-    x_forces : list
-        Dictionary carrying X component of forces with reference to each point on the coupling interface.
-    y_forces : list
-        Dictionary carrying Y component of forces with reference to each point on the coupling interface.
+    forces : list
+        Multiple lists carrying components of forces with reference to each point on the coupling interface.
```
BenjaminRodenberg, line 431: I don't think that this update of the documentation is correct. You are still returning `d` lists below.

fenicsprecice/adapter_core.py
```diff
@@ -427,34 +427,35 @@ def get_forces_as_point_sources(fixed_boundary, function_space, data):
...
-    x_forces = dict()  # dict of PointSources for Forces in x direction
-    y_forces = dict()  # dict of PointSources for Forces in y direction
-
-    fenics_vertices = np.array(list(data.keys()))
+    vertices = np.array(list(data.keys()))
     nodal_data = np.array(list(data.values()))
 
-    # Check for shape of coupling_mesh_vertices and raise Assertion for 3D
-    n_vertices, _ = fenics_vertices.shape
+    n_vertices, dims = vertices.shape
```
BenjaminRodenberg, line 436: Please put an assertion here to ensure that `dims == 2 or dims ==3`.

fenicsprecice/adapter_core.py
```diff
@@ -427,34 +427,35 @@ def get_forces_as_point_sources(fixed_boundary, function_space, data):
...
-    x_forces = filter_point_sources(x_forces, fixed_boundary, warn_duplicate=False)
-    y_forces = filter_point_sources(y_forces, fixed_boundary, warn_duplicate=False)
+    for d in range(dims):
+        forces[d] = filter_point_sources(forces[d], fixed_boundary, warn_duplicate=False)
 
-    return x_forces.values(), y_forces.values()  # don't return dictionary, but list of PointSources
+    if dims == 2:
+        return forces[0].values(), forces[1].values()  # don't return dictionary, but lists of PointSources
+    elif dims == 3:
+        return forces[0].values(), forces[1].values(), forces[2].values()
```
BenjaminRodenberg, line 458: Better return a `d x n` (or `n x d`?) array. I would avoid returning a different number of return values depending on the number of dimensions. This would also resolve the issue above with the documentation.

Additionally: I think this would make it more consistent with the changes introduced in #145.

BenjaminRodenberg (commented):
Also goes into the right direction. Again: I would suggest to directly add a 2D and a 3D unit test for this function.

fenicsprecice/adapter_core.py
IshaanDesai: I am a bit hesitant to change this as it will affect all tutorial cases and in general affect the user-side API of the adapter. Should we still do this now?

fenicsprecice/adapter_core.py
IshaanDesai: Yes I am returning `dims` number of lists so is the description not correct? I mean to say that there are several lists returned. So something like:
```
Multiple lists carrying components of forces with reference to points on the coupling interface.
```

BenjaminRodenberg: The text is fine, but I still find the `forces: list` confusing. Not sure how to actually document multiple return values, if some of them are optional. Maybe something like

```
forces_x : list
    ....
forces_y: list
    ....
forces_z: list
    (optional, only returned, if 3D) ....
```

fenicsprecice/adapter_core.py
BenjaminRodenberg: We should definitely try to keep the API stable. But are you sure? I thought the user only accesses [`precice.get_point_sources`](https://github.com/precice/tutorials/blob/af4282de60f56d8f684cd0caee19acd1f6d69f0f/perpendicular-flap/solid-fenics/solid.py#L187-L188). This function here is in `adapter_core` and hidden from the user. Or am I missing something?

Btw: Do you know that you can unpack return values of a functions: If you `return [first, second]` of `return a_list_of_two` you can receive the return values via `first, second = my_function_returning_a_list_of_two`? Maybe this is a good way of having a consistent return value for different dimensions and at the same time a stable API.

IshaanDesai: The function itself is hidden from the user but we basically return the return of this internal function directly in the following way:
https://github.com/precice/fenics-adapter/blob/f517f2c27b3b2ccf0f0c90359253079ff976833e/fenicsprecice/fenicsprecice.py#L189

hence changing the return statement would inadvertently change the API.

I tried to have unpacked lists in the return statement but it did not work. I will try again.

fenicsprecice/adapter_core.py
```diff
@@ -427,34 +427,34 @@ def get_forces_as_point_sources(fixed_boundary, function_space, data):
...
+            forces[d][key] = PointSource(function_space.sub(d), Point(key), nodal_data[i, d])
 
     # Avoid application of PointSource and Dirichlet boundary condition at the same point by filtering
-    x_forces = filter_point_sources(x_forces, fixed_boundary, warn_duplicate=False)
-    y_forces = filter_point_sources(y_forces, fixed_boundary, warn_duplicate=False)
+    for d in range(dims):
+        forces[d] = filter_point_sources(forces[d], fixed_boundary, warn_duplicate=False)
 
-    return x_forces.values(), y_forces.values()  # don't return dictionary, but list of PointSources
+    return (forces[d].values() for d in range(dims))
```
BenjaminRodenberg, line 457: I think this looks nice now. This does not break the interface, but everything on the developer side is nice and clean.

BenjaminRodenberg (approved):
Looks good :+1: 
Adding some tests would still be great, if you have time.

tests/integration/test_fenicsprecice.py
```diff
@@ -207,103 +207,81 @@ def test_update_expression_vector(self):
...
+        print("dummy_forces Y: {}".format(dummy_forces[1].values()))
+
+        # Get forces from Adapter function
+        forces_x, forces_y = precice.get_point_sources(data)
+
+        print("forces X: {}".format(forces_x))
+        print("forces Y: {}".format(forces_y))
+
+        assert(np.allclose(dummy_forces[0].values(), forces_x))
+        assert(np.allclose(dummy_forces[1].values(), forces_y))
```
IshaanDesai, line 287: @BenjaminRodenberg there is a problem here. The function `get_point_sources` returns a list of PointSource objects. Even if we generate the same PointSource objects it is not possible to check if they carry the same forces as we cannot simple compare objects. Any ideas how we can test this?

Also I think this particular test and also some tests above are in fact unit tests and we keep them in `tests/integration`, which is confusing.

BenjaminRodenberg:
I managed to get the test contributed in 446eebf working. Main problem was that `VectorFunctionSpace` and the scalar rhs `Constant(0)` was not fitting together. I followed chapter 3.3.3 in [the FEniCS tutorials book](https://fenicsproject.org/pub/tutorial/pdf/fenics-tutorial-vol1.pdf) to get this working. Rest were minor issues.

PR #147, opened 2022-01-18
Generalize functions related to extracting vertex information from FEniCS

IshaanDesai:
This PR makes the functions related to extracting vertex information from FEniCS more versatile in terms of its capability to handle problems with different physical dimensions. This PR is one amongst many in the direction of adding the capability of handling 3D cases to the adapter.

### DIFF
--- a/fenicsprecice/adapter_core.py
+++ b/fenicsprecice/adapter_core.py
@@ -218,10 +218,7 @@ def get_fenics_vertices(function_space, coupling_subdomain, dims):
         if coupling_subdomain.inside(v.point(), True):
             lids.append(v.index())
             gids.append(v.global_index())
-            if dims == 2:
-                coords.append([v.x(0), v.x(1)])
-            if dims == 3:
-                coords.append([v.x(0), v.x(1), v.x(2)])
+            coords.append([v.x(d) for d in dims])
 
     return np.array(lids), np.array(gids), np.array(coords)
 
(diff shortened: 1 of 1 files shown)


fenicsprecice/adapter_core.py
```diff
@@ -218,10 +218,7 @@ def get_fenics_vertices(function_space, coupling_subdomain, dims):
         if coupling_subdomain.inside(v.point(), True):
             lids.append(v.index())
             gids.append(v.global_index())
-            if dims == 2:
-                coords.append([v.x(0), v.x(1)])
-            if dims == 3:
-                coords.append([v.x(0), v.x(1), v.x(2)])
+            coords.append([v.x(d) for d in range(dims)])
```
BenjaminRodenberg, line 221: Please add an assertion that `dims == 2 or dims == 3`. Ideally at the beginning of the function.

BenjaminRodenberg (approved):
Looks good :+1: Just a minor observation.

fenicsprecice/adapter_core.py
IshaanDesai: I do not see the need to add an assertion here. This is an internal function of the adapter and is called in `initialize()` where we already check the dimensions. There is no possibility of an arbitrary dimension being passed here.

BenjaminRodenberg: Then let's merge this PR right away ;)

PR #152, opened 2022-05-29
Adding connectivity data for volume coupling

boris-martin:
Added configuration of connectivity data for volume coupling: triangles from the coupling domain are added to the preCICE mesh.
It (seems to) works in serial & parallel, however the parallel version has "holes", as triangles made of owned+unowned vertices are discared:

image

This is however consistent with the current implementation, which does the same in surface coupling with edges.

Questions:
- Currently this is always done. Should it be a parameter, or should we rely on preCICE to ignore useless connectivity? THere are no checks for edges and I went for consistency first...
- There is an assertion that the mesh is made of triangles. Not ideal, but I'm open to suggestions for alternatives. How would it work with quads or in  3D ?
- Some parts of the code aren't very pythonic, I'm open to style suggestions!

Code for testing:

[fenics-adapter-test.tar.gz](https://github.com/precice/fenics-adapter/files/8794241/fenics-adapter-test.tar.gz)

### DIFF
--- a/fenicsprecice/adapter_core.py
+++ b/fenicsprecice/adapter_core.py
@@ -2,7 +2,7 @@
 This module consists of helper functions used in the Adapter class. Names of the functions are self explanatory
 """
 
-from fenics import SubDomain, Point, PointSource, vertices, FunctionSpace, Function, edges
+from fenics import SubDomain, Point, PointSource, vertices, FunctionSpace, Function, edges, cells
 import numpy as np
 from enum import Enum
 import logging
@@ -375,6 +375,8 @@ def get_coupling_boundary_edges(function_space, coupling_subdomain, global_ids, 
         Array of first vertex of each edge.
     vertices2_ids : numpy array
         Array of second vertex of each edge.
+    edges_ids : numpy array
+        Array of FEniCS edge global IDs.
     """
 
     def edge_is_on(subdomain, this_edge):
@@ -386,6 +388,7 @@ def get_coupling_boundary_edges(function_space, coupling_subdomain, global_ids, 
 
     vertices1_ids = []
     vertices2_ids = []
+    edges_ids = []
 
     for edge in edges(function_space.mesh()):
         if edge_is_on(coupling_subdomain, edge):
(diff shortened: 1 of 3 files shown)


BenjaminRodenberg:
Nice to see that there is some improvement in the volume coupling direction :)

What do you mean with "it seems to work in parallel", but at the same time there are holes? If the holes don't have an influence on the result and they are only a visualization artifact, I would not worry too much here.

If there is a general problem with the parallel version, there are two questions I would try to answer before investing too much here: 
1) Do you really need to support parallel runs here? You could also just not support parallel for this case (we also [don't support parallel runs together with PointSources](https://github.com/precice/fenics-adapter/blob/8863cd9e4488da1f52f93749f320ea1a405e2a96/fenicsprecice/fenicsprecice.py#L188), for example) 
2) If you need parallel runs, would it be possible to move to FEniCS-X instead? Moving your example to FEniCS-X might be a big and risky step though, because our FEniCS-X adapter is not stable yet.

The reason I'm asking: (legacy) FEniCS is at its end of life and I've heard that FEniCS-X comes with some improvements w.r.t parallelization. Investing a lot of time into (legacy) FEniCS for a feature that you might not even need might be a bad idea.

boris-martin:
"It seems to work" was just to highlight the fact it worked on my test case but maybe there are edge cases my test didn't cover. But in my case it works 100% as expected. But parallelization does lead to a different results, as elements are missing. 
An analogy in 1D: if you have a splitted domain with a middle vertex as interface, data mapping would do linear interpolation between vertices, except on the edge connecting the boundary, where it would fall back on NN because the boundary isn't part of the preCICE mesh.

So that's not ideal but still better than a pure NN. But the problems lies in the fact that the FEniCS adapter only sets up points that are owned by the rank and not the non-owned shared ones. If we changed that, we could easily add the edges then triangles, but this would require changes out of this PR. (See http://precice.org/couple-your-code-distributed-meshes.html#use-a-single-mesh-and-duplicate-copied-vertices )

I don't personally need parallel runs, I mostly implemented this to keep the best generality possible.

fenicsprecice/adapter_core.py
```diff
@@ -386,19 +389,59 @@ def edge_is_on(subdomain, this_edge):
...
+        assert(len(list(vertices(this_cell))) == 3), "Only triangular meshes are supported"
+        return all([subdomain.inside(v.point(), True) for v in vertices(this_cell)])
+
+    edges_ids = []
 
-    return vertices1_ids, vertices2_ids
+    for cell in cells(function_space.mesh()):
+        if cell_is_in(coupling_subdomain, cell):
+            e1, e2, e3 = list(edges(cell))
+            if e1.index() in precice_edge_dict.keys() and e2.index() in precice_edge_dict.keys() and e3.index() in precice_edge_dict.keys():
```
IshaanDesai, line 441: Where is the variable `precice_edge_dict` defined? Is it somehow a global variable in this file?

fenicsprecice/fenicsprecice.py
```diff
@@ -410,15 +410,20 @@ def initialize(self, coupling_subdomain, read_function_space=None, write_object=
...
                                                                          self._owned_vertices.get_global_ids(),
                                                                          id_mapping)
 
+        # Initialize preCICE edges dict
+        self._precice_edge_dict = {}
+
         for i in range(len(edge_vertex_ids1)):
             assert (edge_vertex_ids1[i] != edge_vertex_ids2[i])
-            self._interface.set_mesh_edge(self._interface.get_mesh_id(self._config.get_coupling_mesh_name()),
+            self._precice_edge_dict[edges_ids[i]] = self._interface.set_mesh_edge(self._interface.get_mesh_id(self._config.get_coupling_mesh_name()),
```
IshaanDesai, line 422: Probably this variables needs to be defined above in the constructor once before use, for consistency.

IshaanDesai:
Some clarification on the points raised in the discussion so far:

> Currently this is always done. Should it be a parameter, or should we rely on preCICE to ignore useless connectivity? THere are no checks for edges and I went for consistency first...

Always passing on connectivity information is fine for now. This would be rectified when https://github.com/precice/fenics-adapter/pull/138 is merged anyway.

> There is an assertion that the mesh is made of triangles. Not ideal, but I'm open to suggestions for alternatives. How would it work with quads or in 3D ?

Lets leave the implementation at 2D for now, anyway the adapter is highly restrictive in 3D. Lets document the restriction of having linear cell interpolation mapping only in 2D. Same goes for the parallelization, I don't think it is worth investing effort in making this work efficiently for parallel cases.

IshaanDesai (approved):

PR #153, opened 2022-07-12
Compatibility update for v3.0.0

BenjaminRodenberg:
Ensures compatibility to precice:develop.


fenicsprecice/fenicsprecice.py
```diff
@@ -251,9 +249,6 @@ def write_data(self, write_function):
         assert (self._write_function_type == determine_function_type(w_func, self._fenics_dims))
         assert (write_function.function_space() == self._write_function_space)
```
BenjaminRodenberg, line 250: Here, we have `assert (...)` (with blank), not `assert(...)` (without blank)

BenjaminRodenberg (commented):
Since the test with the partitioned heat tutorial is failing: I guess we also need to update the tutorials. Into the rabbit hole :rabbit2:

BenjaminRodenberg:
I also added the `v2.0.0` milestone here, because merging this PR would be breaking due to the changes in the action interface. We could also design this in a non-breaking fashion and hide the breaking changes of preCICE `v3` inside of the adapter, but this might be more confusing than useful, I guess.

fenicsprecice/fenicsprecice.py
IshaanDesai: Every `assert` is now of the style `assert (...)`.

BenjaminRodenberg: All fine. As long as it is consistent :)

fenicsprecice/expression_core.py
```diff
@@ -128,10 +128,10 @@ def is_scalar_valued(self):
         """
         try:
             if self._vals.ndim == 1:
-                assert(self._function_type is FunctionType.SCALAR)
+                assert (self._function_type is FunctionType.SCALAR)
```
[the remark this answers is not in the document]
BenjaminRodenberg: I can live with both, as long as it is consistently enforced :)

IshaanDesai (approved):

BenjaminRodenberg:
Since this PR is already quite old: At the moment it is not on the first place of my list of priorities, but I expect to work with preCICE v3 + FEniCS within the next few months again. We can wait with this issue until then and I will occasionally review and merge. If something else pops up in the meantime, we can, of course, take care of this issue earlier.

IshaanDesai:
> Since this PR is already quite old: At the moment it is not on the first place of my list of priorities, but I expect to work with preCICE v3 + FEniCS within the next few months again. We can wait with this issue until then and I will occasionally review and merge. If something else pops up in the meantime, we can, of course, take care of this issue earlier.

I plan to work on this towards the end of this month. Most of the updates in the adapter itself have been done. The major work is actually in updating the tutorials in https://github.com/precice/tutorials/pull/333.

BenjaminRodenberg:
@IshaanDesai I'll try to wrap this PR up soon. Main problem here are the unit tests. I assume the tutorial tests should work as soon as https://github.com/precice/tutorials/pull/333 is merged.

fenicsprecice/adapter_core.py
```diff
@@ -415,31 +412,36 @@ def get_coupling_triangles(function_space, coupling_subdomain, precice_edge_dict
...
             e1, e2, e3 = list(edges(cell))
-            if all(edge in precice_edge_dict.keys() for edge in [e1.index(), e2.index(), e3.index()]):
-                edges_ids.append([e1.index(), e2.index(), e3.index()])
-
-    return np.array(edges_ids)
+            if all(edge_ids in fenics_edge_ids for edge_ids in [e1.index(), e2.index(), e3.index()]):
+                v1, v2 = vertices(e1)
+                _, v3 = vertices(e2)
+                assert (v3 != v1)
+                assert (v2 != v2)
```
BenjaminRodenberg, line 439: I found this problem by running `channel-transport-reaction`. But even fixing this obviously wrong assertion does not help. There seems to be some issue with the function `get_coupling_triangles`. We obviously need a test for this function and need to fix it. @IshaanDesai can you take a look?

PR #161, opened 2023-08-02
Build `latest` docker container from master not `develop` branch

### LINKED ISSUE #160: precice/fenics-adapter:latest is not stable but the develop release
Contrary to expectation the docker container published with tag `latest` seems to be the same as `develop`. 
It should however use the main/master to build. 
We need this to have a stable version compatible with precicev2 to test in the system tests.

valentin-seitz:
Should fix #160 
Here a similar idea to the python-bindings applies: 
I Made the Workflow a bit more usable via the manual dispatch and also split up `latest` and `develop` tags accordingly. 

Additionally the Dockerfile is more in line with the current approach to Dockerfiles used in the Systemtestsv2

### DIFF
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -24,11 +24,15 @@ jobs:
       - name: Set branch name for manual triggering
         if: github.event_name == 'workflow_dispatch'
         shell: bash
-        run: echo "branch=${{ inputs.branch }}" >> $GITHUB_ENV    
+        run: | 
+          echo "ADAPTER_REF=${{ inputs.branch }}" >> $GITHUB_ENV
+          echo "PYTHON_REF=latest" >> $GITHUB_ENV       
       - name: Set branch name for on pull triggering
         if: github.event_name != 'pull_request' && github.event_name != 'workflow_dispatch'
         shell: bash
-        run: echo "branch=${{ github.ref_name }}" >> $GITHUB_ENV    
+        run: |
+          echo "ADAPTER_REF=${{ github.ref_name }}" >> $GITHUB_ENV
+          echo "PYTHON_REF=develop" >> $GITHUB_ENV          
       - name: Checkout Repository
         uses: actions/checkout@v2
       - name: Set up Docker Buildx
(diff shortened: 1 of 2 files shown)


valentin-seitz:
I dont know about the checks: The Linting complains about missing links. So for me it seems that its not related to this PR.

MakisH (approved):
The changes make sense to me. :+1:

I guess this is still draft as it depends on https://github.com/precice/python-bindings/pull/186

> I dont know about the checks: The Linting complains about missing links. So for me it seems that its not related to this PR.

They are indeed unrelated (and need to be updated). The tests also fail because the configurations have yet to be ported to preCICE v3.

@IshaanDesai I let you do the merging.

valentin-seitz:
I just created the merge to master because the the workflows that are "normally" triggered are the ones available on master. 
But this could go to develop when the workflow is not started now from master and therefore overriding the image with the`latest` tag again .

MakisH:
I understand that there is no reason (anymore) to target `master`, so let's just change the target branch to `develop` and merge.

@valentin-seitz you can change the target branch by clicking "Edit" on the top.

IshaanDesai:
I pointed the PR to develop. It is still draft, but is this ready for review and merging?

valentin-seitz:
> I pointed the PR to develop. It is still draft, but is this ready for review and merging?

This is indeed ready for merging. 
Note that similar to the python-bindings, this will change will not be effective until its merged to master.

IshaanDesai:
> Note that similar to the python-bindings, this change will not be effective until its merged to master.

Do we need this to work before version 3 of all downstream packages are released?

valentin-seitz:
> > Note that similar to the python-bindings, this change will not be effective until its merged to master.
> 
> Do we need this to be effective before version 3 of all downstream packages are released?

From my perspective: not really. I currently rely on the `latest` tag to be compatible with v2 but i can easily re trigger the publishing of the image with the `latest` tag manually

MakisH:
We don't need to update `master`, but I guess we can already merge this, right? If yes, @IshaanDesai I leave it up to you.

IshaanDesai (approved):

PR #168, opened 2024-02-16
Disable build-docker workflow on push to master

BenjaminRodenberg:
See https://github.com/precice/python-bindings/pull/196. We decided to not release a docker container for the python bindings and the fenics adapter, because this is unnecessary.


IshaanDesai (approved):

MakisH (approved):
Indeed, not necessary anymore for the system tests.

PR #170, opened 2024-03-09
Support listlike checkpoints

### LINKED ISSUE #169: Support checkpointing of multiple fields
Currently, the adapter only supports checkpointing for a single field. It would be useful to be able to checkpoint an arbitrary number of fields like suggested in https://github.com/precice/tutorials/issues/50.

This feature was also requested in our forum. See https://precice.discourse.group/t/fenics-precice-implicit-coupling-with-multiple-fields-in-checkpoint/1806.

Note: We need to implement this in a backwards compatible and non-breaking fashion. Meaning, if we follow the approach suggested in https://github.com/precice/tutorials/issues/50 the checkpointing functions need to be able to accept `u_n` and `(u_n)` (no tuple vs. tuple).

BenjaminRodenberg:
Closes #169

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## latest
 
-* Additionally support checkpoints being provided as a list or tuple (of FEniCS Functions). https://github.com/precice/fenics-adapter/pull/170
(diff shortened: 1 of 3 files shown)


IshaanDesai (approved):
Looks reasonable :+1:

PR #173, opened 2024-08-03
Tests checking implementation of checkpointing/SolverState

NiklasVin:
These tests are written in context of Pull Request https://github.com/precice/fenics-adapter/pull/172
Since the core functionality of the checkpointing is implemented in the SolverState class, the tests are actually testing only this class. 
This commit consists of three different tests:
1. Write checkpoint & directly load it (just for the sake of completeness)
2. Write checkpoint of a function, then modify the dof vector of the dolfin function by directly accessing it via `vector()`. After that, load the checkpoint.
3. Write checkpoint of a function, then modify the this dolfin function via `assign()`. After that, load the checkpoint.

### DIFF
--- a/tests/unit/test_checkpointing.py
+++ b/tests/unit/test_checkpointing.py
@@ -1,7 +1,6 @@
 from unittest.mock import MagicMock
 from unittest import TestCase
 from fenics import FunctionSpace, UnitSquareMesh, Expression, interpolate
-import random as rnd
 from fenicsprecice.solverstate import SolverState
 
 
--- a/tests/unit/test_checkpointing.py
+++ b/tests/unit/test_checkpointing.py
@@ -39,25 +39,26 @@ class TestCheckpointing(TestCase):
         size = 5
         mesh = UnitSquareMesh(size,size)
         V = FunctionSpace(mesh, 'P', 2)
-        dummy_value = 1
-        E = Expression("t", t=dummy_value, degree=2)
+        ref_value = 1
+        E = Expression("t", t=ref_value, degree=2)
         u = interpolate(E, V)
 
         # "write checkpoint"
-        sstate = SolverState(u, dummy_value, n)
+        sstate = SolverState(u, ref_value, n)
 
         # modify state of u
-        u.vector()[:] = dummy_value + 2
+        dummy_value = ref_value + 2
+        u.vector()[:] = dummy_value
 
         # "read checkpoint"
         u_cp, _, _ = sstate.get_state()
 
         #check values
         #function should be the same everywhere
-        #(so the vector values should all be dummy_value=1)
+        #(so the vector values should all be ref_value)
         vec_u_cp = u_cp.vector()
         for i in range(size*size):
-            self.assertAlmostEqual(dummy_value, vec_u_cp[i])
+            self.assertAlmostEqual(ref_value, vec_u_cp[i])
     
 
     def test_solverstate_modification_assign(self):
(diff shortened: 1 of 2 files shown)


BenjaminRodenberg:
To make sure: At the moment we expect this test to fail. As soon as https://github.com/precice/fenics-adapter/pull/172 is merged this test should succeed. Correct?

tests/unit/test_checkpointing.py
```diff
@@ -0,0 +1,93 @@
...
+from fenicsprecice.solverstate import SolverState
+
+
+class TestCheckpointing(TestCase):
+    def test_solverstate_basic(self):
+        """
+        Check if correct values are read from the checkpoint, while the state of the object that is copied is not changed
+        """
+        time = 0.5
+        iteration = 1
```
BenjaminRodenberg, line 13:
```suggestion
        n = 1
```
why not `n` to make it consistent with `n_cp`? Same for below.

tests/unit/test_checkpointing.py
```diff
@@ -0,0 +1,93 @@
...
+        size = 5
+        mesh = UnitSquareMesh(size,size)
+        V = FunctionSpace(mesh, 'P', 2)
+        E = Expression("t", t=time, degree=2)
```
BenjaminRodenberg, line 17: Why use `time` here? I think this makes it unnecessarily hard to read. Why not the following:

```suggestion
        dummy_value = 42
        E = Expression("x", x=dummy_value, degree=2)
```

tests/unit/test_checkpointing.py
```diff
@@ -0,0 +1,93 @@
...
+        u = interpolate(E, V)
+
+        # "write checkpoint"
+        sstate = SolverState(u, time, iteration)
+
+        # modify state of u
+        u.vector()[:] = time + 2
```
BenjaminRodenberg, line 50: See above. I think the following would be much easier to read:

```suggestion
        dummy_value += 2
        u.vector()[:] = dummy_value
```

BenjaminRodenberg (changes requested):
Some minor readability concerns. As soon as you merged the current state of `develop` into this PR the tests should work again.

NiklasVin:
> To make sure: At the moment we expect this test to fail. As soon as #172 is merged this test should succeed. Correct?

Yes

tests/unit/test_checkpointing.py
NiklasVin: I would suggest to leave it like that, since the assertion still needs the original `dummy_value`. Then, I would need to subtract the 2 in the assertion instead.

BenjaminRodenberg: Then why not call it `reference_value` and `dummy_value`? You could then put a `reference_value = dummy_value` before you start modifying `dummy_value`.

But if you want to leave it as it is: Also ok.

BenjaminRodenberg:
Ok. Then I will merge the tests now. This leads to failure on `develop` but then we have #172 as a fix.

tests/unit/test_checkpointing.py
NiklasVin: I saw this comment just now. I added the changes you suggested.

BenjaminRodenberg: Looks good.

BenjaminRodenberg (approved):
Looks good to me and also tests the right thing! Thanks for the contribution. Will merge this now.

PR #178, opened 2024-08-29
Make t and n optional arguments in store_checkpoint

### LINKED ISSUE #73: Make t and n optional in checkpointing
See https://github.com/precice/fenics-adapter/pull/59#issuecomment-635939898

NiklasVin:
Closes https://github.com/precice/fenics-adapter/issues/73.


BenjaminRodenberg (changes requested):
Since I expect that this is a quite niche feature: A test would be important.

Additionally: In https://github.com/precice/fenics-adapter/pull/59#issuecomment-635939898 I mention some tutorials where we could just skip `t` and/or `n`. Let's prepare a PR where we actually implement this in the tutorials. This simplifies the tutorials and also acts as an integration test for this feature.

fenicsprecice/fenicsprecice.py
```diff
@@ -459,14 +459,25 @@ def retrieve_checkpoint(self):
...
+        payload, t, n = self._checkpoint.get_state()
+        match (t, n):
+            case (None, None):
+                return payload
+            case (_, None):
+                return payload, t
+            case (None, _):
+                return payload, n
+            case _:
+                return payload, t, n
```
BenjaminRodenberg: I just tested this PR with the perpendicular flap. Generally, it seems to work. But I'm not so happy with the implications on the return value of the `retrieve_checkpoint` function: If I call
```
cp, t_or_n = adapter.retrieve_checkpoint()
```
the second argument could be anything. The actual return value depends on how I called `store_checkpoint` before. This does not look right to me. Sorry to bring this up so late.

## Non-breaking solution

As a non-breaking solution, I can only imagine returning `None`, if `t` or `n` has not been provided previously. This results in always calling
```
cp, t, n = adapter.retrieve_checkpoint()
```
this would be independent from what we did with `store_checkpoint` before.

## Breaking solutions

I think a more reasonable way would be the following:
```
cp = adapter.retrieve_checkpoint()
```
and then
```
value = cp.value
t = cp.t
n = cp.n
```
Meaning: `cp` would be a named tuple (see https://stackoverflow.com/questions/2970608/what-are-named-tuples-in-python). Scipy uses something similar in `integrate.solve_ivp` and calls it a "bunch object" (see https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.solve_ivp.html).

This brings us to the next question: Why not just use the following API:
```
store_checkpoint(self, payload: namedtuple)

retreive_checkpoint() -> namedtuple
```
We would then have to allow for FEniCS Functions or simple float or int values in the `payload`. But this would be an implementation detail and the user just gets back whatever was provided in the original payload.

Important about both proposed solutions: They would be breaking (and therefore, we can only release them with fenicsprecice 3.0.0).

## What to do

I would suggest to implement and merge the non-breaking solution. We can open an issue describing the breaking solution and schedule it for version `3.0.0` (probably this will happen not too soon).

BenjaminRodenberg (changes requested):

BenjaminRodenberg (approved):
Looks good 👍

Issue #180, opened 2024-09-04
`mpi4py` version 4.0.0 prevents installation of FEniCS adapter


NiklasVin:
Legacy FEniCS is not compatible with version 4.0.0 of `mpi4py`. During installation of the FEniCS adapter, the FEniCS installation is tested.
<details>

<summary> Therefore, the FEniCS adpater cannot be installed, as FEniCS produces an error. </summary>

```  
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [22 lines of output]
      ERROR: could not import mpi4py!
      ValueError: mpi4py.MPI.Op size changed, may indicate binary incompatibility. Expected 56 from C header, got 40 from PyObject
      
      The above exception was the direct cause of the following exception:
      
      Traceback (most recent call last):
        ...
          from fenics import *
        File "/usr/lib/python3/dist-packages/fenics/__init__.py", line 7, in <module>
          from dolfin import *
        File "/usr/lib/petsc/lib/python3/dist-packages/dolfin/__init__.py", line 144, in <module>
          from .fem.assembling import (assemble, assemble_system, assemble_multimesh, assemble_mixed,
        File "/usr/lib/petsc/lib/python3/dist-packages/dolfin/fem/assembling.py", line 34, in <module>
          from dolfin.fem.form import Form
        File "/usr/lib/petsc/lib/python3/dist-packages/dolfin/fem/form.py", line 12, in <module>
          from dolfin.jit.jit import dolfin_pc, ffc_jit
        File "/usr/lib/petsc/lib/python3/dist-packages/dolfin/jit/jit.py", line 124, in <module>
          def compile_class(cpp_data, mpi_comm=MPI.comm_world):
                                               ^^^^^^^^^^^^^^
      RuntimeError: Error when importing mpi4py
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed
```

</details>

This issue is also related to https://github.com/precice/tutorials/pull/563 and https://github.com/precice/precice/pull/2052#issuecomment-2321693938.

PR #181, opened 2024-09-04
Modify `setup.py` to avoid crash due to version 4.0.0 of `mpi4py` during installation

NiklasVin:
This PR closes https://github.com/precice/fenics-adapter/issues/180.

`setup_requires=['mpi4py<4']` is necessary as this requirement is needed for the imported FEniCS below.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,6 @@ setup(name='fenicsprecice',
       author_email='info@precice.org',
       license='LGPL-3.0',
       packages=['fenicsprecice'],
-      setup_requires=['mpi4py<4'],
       install_requires=['pyprecice>=3.0.0.0', 'scipy', 'numpy>=1.13.3, <2', 'mpi4py<4'],
       test_suite='tests',
       zip_safe=False)
--- a/setup.py
+++ b/setup.py
@@ -14,15 +14,6 @@ if sys.version_info[1] == 6 and sys.version_info[2] == 9:
                   "when running the tests (see https://github.com/precice/fenics-adapter/pull/61). If you want to "
                   "run the tests, please install a different Python version.")
 
-try:
-    from fenics import *
-except ModuleNotFoundError:
-    warnings.warn("No FEniCS installation found on system. Please install FEniCS and check the installation.\n\n"
-                  "You can check this by running the command\n\n"
-                  "python3 -c 'from fenics import *'\n\n"
-                  "Please check https://fenicsproject.org/download/ for installation guidance.\n"
-                  "The installation will continue, but please be aware that your installed version of the "
-                  "fenics-adapter might not work as expected.")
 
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -39,6 +30,17 @@
       author_email='info@precice.org',
       license='LGPL-3.0',
       packages=['fenicsprecice'],
-      install_requires=['pyprecice>=3.0.0.0', 'scipy', 'numpy>=1.13.3, <2', 'mpi4py'],
+      setup_requires=['mpi4py<4'],
```
BenjaminRodenberg, line 33: I'm not sure what `setup_requires` exactly does but I would not put a hard requirement here for the sake of the check that we have below. If the check does not succeed there is probably a good reason for the warning. We should not try to fix a broken FEniCS installation with this `setup.py`.

setup.py
```diff
@@ -39,6 +30,17 @@
...
+
+try:
+    from fenics import *
+except ModuleNotFoundError:
+    warnings.warn("No FEniCS installation found on system. Please install FEniCS and check the installation.\n\n"
+                  "You can check this by running the command\n\n"
+                  "python3 -c 'from fenics import *'\n\n"
+                  "Please check https://fenicsproject.org/download/ for installation guidance.\n"
+                  "The installation will continue, but please be aware that your installed version of the "
+                  "fenics-adapter might not work as expected.")
```
BenjaminRodenberg, line 46: See above. Try not to fix a broken FEniCS installation with this `setup.py` then you can also move this up to the original position. Note: We might remove this warning very soon anyways.

BenjaminRodenberg (changes requested):
Thanks for opening this PR. Some comments. I, unfortunately, have no idea why the CI is currently failing...

setup.py
BenjaminRodenberg: I looked a bit into the reason why `mpi4py` is failing and actually we are exactly trying to do this: We are fixing a broken FEniCS installation and this is also what we need to do because I think we cannot expect that there will be a corresponding fix for (legacy) FEniCS (either supporting `mpi4py>=4` or restricting to `mpi4py<4`).

BenjaminRodenberg:
I removed the checks. See https://github.com/precice/fenics-adapter/pull/182.

NiklasVin:
@BenjaminRodenberg So we can close this PR since you removed the checks, right?

BenjaminRodenberg:
> @BenjaminRodenberg So now, we only need to require
> 
> ```diff
> + install_requires=['pyprecice>=3.0.0.0', 'scipy', 'numpy>=1.13.3, <2', 'mpi4py<4']
> ```
> 
> now?

I hope so.

BenjaminRodenberg (approved):
I synchronized this PR and removed the `setup_requires`. CI-wise everything looks good and I will merge this PR latest on the weekend.

Issue #179, opened 2024-09-03
Introduce `pyproject.toml`


BenjaminRodenberg:
It is recommended to use `pyproject.toml`.

I would suggest to start by managing the sollowing additional points over `pyproject.toml`:

* Dependencies
    * `mpi4py < 4` is required. See https://github.com/precice/tutorials/pull/563
    * Add `fenics` as a dependency? Note that the installation process of FEniCS is somewhat non-standard. See, for example, what I'm doing [here](https://github.com/precice/tutorials/blob/0225e428cc559444a29541b9c46318d8c60cd8fd/partitioned-heat-conduction/solver-fenics/requirements.txt#L5-L11)
* According to https://ianhopkinson.org.uk/2022/02/understanding-setup-py-setup-cfg-and-pyproject-toml-in-python/ we can also replace the `setup.cfg` file with `pyproject.toml`. I introduced `setup.cfg` to be able to use `versioneer`. We should be able to move that to `pyproject.toml` (see https://github.com/python-versioneer/python-versioneer?tab=readme-ov-file#vendored-mode).
* Remove the python version checks from `setup.py` and into `pyproject.toml`. See https://stackoverflow.com/questions/76161889/how-do-i-specify-a-python-version-constraint-for-a-dependency-specifically-for-a

fsimonis:
https://hatch.pypa.io/latest/intro/#existing-project
This can convert the setup.py to pyproject.toml

IshaanDesai:
Done via https://github.com/precice/fenics-adapter/pull/204

PR #182, opened 2024-09-05
Remove checks for FEniCS and python3

BenjaminRodenberg:
I removed the checks for the following reasons:

1. FEniCS check turns out problematic in #181 
2. If we follow the recommended installation procedure `pip install .` the warnings are not shown. The warnings only appear for the deprecated `python3 setup.py install`

We can check for the python version (and potentially also FEniCS) when moving to `pyproject.toml` in a better way (see #179).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,8 +1,6 @@
 import os
 from setuptools import setup
 import versioneer
-import warnings
-
 this_directory = os.path.abspath(os.path.dirname(__file__))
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -3,27 +3,6 @@
 import versioneer
 import warnings
```
fsimonis, line 5: This can go too.

```suggestion

```

fsimonis (approved):

Issue #164, opened 2023-10-02
recommended install fails under Ubuntu


ng0177:
This error appears under Ubuntu 23.04
```
$ pip3 install --user fenicsprecice
error: externally-managed-environment
```

BenjaminRodenberg:
@IshaanDesai I did not follow the discussion on discourse. Are there any updates or solutions we could offer here?

BenjaminRodenberg:
@NiklasVin I think this is mainly a documentation issue that we can resolve by updating the `README.md`. For more recent `pip` version it is generally recommended to use `venv` (we are introducing the corresponding changes in the tutorials currently https://github.com/precice/tutorials/issues/547)

PR #183, opened 2024-09-15
Clarify installation procedure

NiklasVin:
Closes https://github.com/precice/fenics-adapter/issues/164


BenjaminRodenberg (approved):
I moved the instructions to a different location and added the commands that we use in the tutorials. I also removed the `--user` flag because it is not needed anymore for the recommended approach with venvs.

PR #184, opened 2024-10-25
Release v2.2.0

BenjaminRodenberg:
This release comes with some minor modifications to simplify the installation process and an important bugfix when subcycling is used (see #172).

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -22,13 +22,13 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install setuptools wheel twine
+          pip3 install --break-system-packages setuptools wheel twine
       - name: Build distribution
         run:  python3 setup.py sdist
   test:
     name: Run mock unit tests
     runs-on: ubuntu-latest
-    container: quay.io/fenicsproject/stable:current
+    container: benjaminrodenberg/fenics
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v2      
@@ -37,7 +37,7 @@ jobs:
           mkdir -p precice
           echo "from setuptools import setup" >> precice/setup.py
           echo "setup(name='pyprecice', version='3.0.0.0')" >> precice/setup.py
-          python3 -m pip install ./precice/
+          pip3 install --break-system-packages ./precice/
       - name: Run unit tests
         run:  python3 setup.py test -s tests.unit
       - name: Run integration tests
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Use `copy(deepcopy=True)` when checkpointing to make checkpointing more user-friendly and secure. IMPORTANT: might increase runtime, please open an issue if you face serious problems. [#172](https://github.com/precice/fenics-adapter/pull/172)
 * Add unit tests for checkpointing. [#173](https://github.com/precice/fenics-adapter/pull/173)
+* Remove checks for FEniCS installation and python3 from `setup.py` since the approach is deprecated. [#182](https://github.com/precice/fenics-adapter/pull/182)
 
 ## 2.1.0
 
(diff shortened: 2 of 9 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -22,13 +22,13 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install setuptools wheel twine
+          pip3 install --break-system-packages setuptools wheel twine
```
IshaanDesai, line 25: Why not use pipx and do things cleanly here?

.github/workflows/build-and-test.yml
```diff
@@ -37,7 +37,7 @@ jobs:
           mkdir -p precice
           echo "from setuptools import setup" >> precice/setup.py
           echo "setup(name='pyprecice', version='3.0.0.0')" >> precice/setup.py
-          python3 -m pip install ./precice/
+          pip3 install --break-system-packages ./precice/
```
IshaanDesai, line 40: Same as above.

IshaanDesai (changes requested):
Code related changes make sense; the CI stuff not so much.

.github/workflows/build-and-test.yml
```diff
@@ -22,13 +22,13 @@ jobs:
...
           python -m pip install --upgrade pip
-          pip install setuptools wheel twine
+          pip3 install --break-system-packages setuptools wheel twine
       - name: Build distribution
         run:  python3 setup.py sdist
   test:
     name: Run mock unit tests
     runs-on: ubuntu-latest
-    container: quay.io/fenicsproject/stable:current
+    container: benjaminrodenberg/fenics
```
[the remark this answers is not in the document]
BenjaminRodenberg: Currently, we do not have a `precice/fenics` image and I did not want to get into the hassle creating one. `quay.io/fenicsproject/stable:current` does not work anymore. See https://github.com/precice/fenics-adapter/pull/174.

.github/workflows/build-and-test.yml
BenjaminRodenberg: I have not used `pipx` a lot but I think `pipx` is not the right tool here. If I try it locally I get the following error:

```
$ pipx install setuptools

No apps associated with package setuptools or its dependencies. If you are
attempting to install a library, pipx should not be used. Consider using pip
or a similar tool instead.
```

I think the alternatives to `pip` are installation through the package manager, i.e., `apt install`, or going the `venv` path. Sure, we could do this, but I do not see a great benefit and for the sake of testing this approach is also good enough.

If you can suggest a one-liner change with `pipx`, ok. But I would rather not change more and keep it as it is. We can still change it with the next release.

.github/workflows/build-and-test.yml
BenjaminRodenberg: Same as above :smirk:

NiklasVin:
There are also two open PRs I created (#178 and #183) that close two issues (from 2020 and 2023). Don't we want to add them to the new release as well? If releasing this version is urgent, it makes sense to not include them.

.github/workflows/build-and-test.yml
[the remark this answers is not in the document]
IshaanDesai: By not having the image in the precice account, I see a maintenance liability problem here which will hurt us in the future. But if it is too much work right now, we can keep this as it is.

.github/workflows/build-and-test.yml
IshaanDesai: I tried to use `apt` to install the necessary build dependencies: https://github.com/precice/fenics-adapter/pull/185 but this does not seem to work. I do not want to invest too much time here, so lets go with what we have for now.

IshaanDesai (approved):
My earlier comments have been addressed to a reasonable degree. Release is good to go :+1:

.github/workflows/build-and-test.yml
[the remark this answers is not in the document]
BenjaminRodenberg: Opened an issue: #186

Issue #189, opened 2025-01-21
Implement testing in CI pipeline using venv


BenjaminRodenberg:
Due to legacy developments our tests in the CI pipeline are executed relying on system dependencies. This approach is problematic:

* [Here](https://github.com/precice/fenics-adapter/blob/e9fbeb3f15270525f83effca984f3e90befc25be/.github/workflows/build-and-test.yml#L40) we need to use the option `--break-system-packages` in order to install the fake preCICE package
* In https://github.com/precice/fenics-adapter/commit/e9fbeb3f15270525f83effca984f3e90befc25be the tests don't work since SciPy is not installed on the system. I manually installing SciPy via `apt install python3-scipy` fixes this issue and I'm planning to add this as a work around soon.

Note that working on #177 might also fix these problems. But I'm not very familiar with `tox` and some additional steps might be necessary. I am therefore opening this issue for a better separation of concerns.

BenjaminRodenberg:
#197 closes this PR.

Issue #177, opened 2024-08-20
Use `tox` for testing


BenjaminRodenberg:
We currently use `python3 setup.py test` which is already deprecated for quite some time. I, therefore, get the following warning when running tests via `python3 setup.py test`:

```
WARNING: Testing via this command is deprecated and will be removed in a future version. Users looking for a generic test entry point independent of test runner are encouraged to use tox.
```

BenjaminRodenberg:
I will probably start working on this issue with a student visiting our chair end of March

BenjaminRodenberg:
Note that for newer versions of `setuptools` the command `python3 setup.py test` actually is not supported anymore. After installing `setuptools==72` via `pip3 install setuptools==72` I get the following:

```
fenics-adapter$ python3 setup.py test
...
error: invalid command 'test'
```

Downgrading to `setuptools==71.1.0` via `pip3 install setuptools==71.1.0` fixes the issue and `python3 setup.py test` works as expected.

This clearly shows that we should move from `python3 setup.py test` to something more modern.

PR #190, opened 2025-01-21
Manually install scipy in tests

BenjaminRodenberg:
The goal of this fix is to bring the test back to a working state. On the long run the plan is to modernize the testing infrastructure (see #189 and #177).

The approach implemented in this PR is problematic. Ideally the python package should define the dependency on scipy on its own and the installation of scipy should happen automatically.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -32,8 +32,8 @@ jobs:
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v2   
-      - name: Update apt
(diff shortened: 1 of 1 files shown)


BenjaminRodenberg:
@IshaanDesai and @NiklasVin: This is meant mainly as an FYI. The solution is temporary and I will probably overhaul the testing at a later point in time and bring everything to a more modern infrastructure.

IshaanDesai (approved):
Looks good (as a temporary fix) :+1:

PR #191, opened 2025-01-22
Install FEniCS via `apt-get`

BenjaminRodenberg:
Instead of using (and maintaining) the container `benjaminrodenberg/fenics` we can also use a github action to install FEniCS. As far as I understand the action automatically uses a caching mechanism to keep the time for installing packages on top of the base container at a minimum.


BenjaminRodenberg:
Interesting observation: The caching I mentioned above seems to cause some problem. The first time I installed FEniCS using the action everything worked. The second time (when the cache is used) it does not work anymore. I reran the tests that originally worked and they also seem to be broken now (presumably due to caching).

I will now go without the action as an alternative without caching.

BenjaminRodenberg:
There is no huge difference in runtime for using a docker container with pre-installed FEniCS vs. installing it in the test via `apt-get`. I would suggest merging this PR since it avoid the additional effort for maintaining the docker container. It also simplifies the testing since we have one less component to worry about.

IshaanDesai (approved):
This is indeed a cleaner and more sustainable solution :+1:

PR #195, opened 2025-03-06
Restrict SciPy version

### LINKED ISSUE #194: Tests fail for `SciPy` version `>=1.15.0`
The tests for the adapter fail with a SciPy version that is greater than 1.15.0 with the error `ModuleNotFoundError: No module named 'scipy.optimize._highspy._core.simplex_constants'; 'scipy.optimize._highspy._core' is not a package`.

NiklasVin:
A temporary workaround for https://github.com/precice/fenics-adapter/issues/194.

*edit* Closes #194

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # FEniCS-preCICE adapter changelog
 
+## develop
+
+* Set required version of `scipy` to `<1.15.0` to prevent failing tests because of a required module that is not found for later versions of scipy. [#195](https://github.com/precice/fenics-adapter/pull/195)
+
 ## v2.2.0
 
 * Use `copy(deepcopy=True)` when checkpointing to make checkpointing more user-friendly and secure. IMPORTANT: might increase runtime, please open an issue if you face serious problems. [#172](https://github.com/precice/fenics-adapter/pull/172)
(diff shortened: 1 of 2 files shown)


BenjaminRodenberg:
We already discussed this personally: I would suggest to make #194 a `wontfix` and just live with the fact that SciPy 1.15 and greater is not supported by the adapter since we are planning to move to FEniCSx sooner or later..

BenjaminRodenberg (approved):
Works for me. Thank you!

One final todo: Please add a new line in `CHANGELOG.md`.

MakisH:
@NiklasVin is this still necessary?

Asking because now installing the adapter on Ubuntu 26.04 takes significant time to build the older scipy.

NiklasVin:
@MakisH Sorry for the late reply. To be honest, I don't know if this is necessary for a working adapter. If I see it correctly, the restriction was necessary because some tests failed with more recent versions of scipy. However, I don't know whether this affects the functionality of the adapter itself.

MakisH:
@NiklasVin then, I would suggest loosening the restriction and see if the tests now pass. I can prepare a PR, unless you prefer to.

IshaanDesai:
> I can prepare a PR, unless you prefer to.

Please go ahead :)

PR #197, opened 2025-03-11
Introduce tox for testing

### LINKED ISSUE #177: Use `tox` for testing

giuliatum:
Closes https://github.com/precice/fenics-adapter/issues/177

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -33,9 +33,15 @@ jobs:
         uses: actions/checkout@v2   
       - name: Install FEniCS and scipy (see https://github.com/precice/fenics-adapter/issues/189)
         run:  |
-          sudo add-apt-repository ppa:fenics-packages/fenics
+          sudo add-apt-repository ppa:fenics-packages/fenics -y
           sudo apt-get update
-          sudo apt-get install -y python3-scipy fenics
+          sudo apt-get install -y \ 
+            python3-scipy \ 
+            fenics \
+            pkg-config \
+            libprecice-dev \
+            build-essential
+          pkg-config --modversion precice
       - name: Install tox
         run: pip install tox
       - name: Run unit tests
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -33,15 +33,18 @@ jobs:
         uses: actions/checkout@v2   
       - name: Install FEniCS and scipy (see https://github.com/precice/fenics-adapter/issues/189)
         run:  |
-          sudo add-apt-repository ppa:fenics-packages/fenics -y
-          sudo apt-get update
-          sudo apt-get install -y \ 
-            python3-scipy \ 
-            fenics \
-            pkg-config \
-            libprecice-dev \
-            build-essential
-          pkg-config --modversion precice
+         apt-get update -y
+         apt-get install -y software-properties-common  # Required for add-apt-repository
+         add-apt-repository ppa:fenics-packages/fenics -y
+         apt-get update -y
+         apt-get install -y \
+          python3-scipy \
+          fenics \
+          pkg-config \
+          libprecice-dev \
+          build-essential
+         pkg-config --modversion precice
+         python3 -c "import fenics; print(f'FEniCS {fenics.__version__}')"
       - name: Install tox
         run: pip install tox
       - name: Run unit tests
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -42,6 +42,6 @@ jobs:
           echo "setup(name='pyprecice', version='3.0.0.0')" >> precice/setup.py
           pip3 install --break-system-packages ./precice/
       - name: Run unit tests
-        run:  python3 setup.py test -s tests.unit
+        run:  tox run -e unit
       - name: Run integration tests
-        run:  python3 setup.py test -s tests.integration
+        run:  tox run -e integration
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -40,8 +40,7 @@ jobs:
           mkdir -p precice
           echo "from setuptools import setup" >> precice/setup.py
           echo "setup(name='pyprecice', version='3.0.0.0')" >> precice/setup.py
-          pip3 install --break-system-packages ./precice/
-          name: Install tox
+          pip3 install --break-system-packages ./precice
       - name: Install tox
         run: pip install tox
       - name: Run unit tests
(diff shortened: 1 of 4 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -2,8 +2,7 @@ name: Build and Test
 on:
   push:
     branches:
-      - master
-      - develop
+      - '*'
```
BenjaminRodenberg, line 5: Let's revert this before merging but fine for development purposes.

setup.py
```diff
@@ -16,6 +16,6 @@
       author_email='info@precice.org',
       license='LGPL-3.0',
       packages=['fenicsprecice'],
-      install_requires=['pyprecice>=3.0.0.0', 'scipy<1.15.0', 'numpy>=1.13.3, <2', 'mpi4py<4'],
```
BenjaminRodenberg: We also have to revert this before merging but ok during development to avoid having to install preCICE on the system.

.github/workflows/build-and-test.yml
```diff
@@ -2,8 +2,7 @@ name: Build and Test
...
   pull_request:
     paths:
       - '**'
```
BenjaminRodenberg:
```suggestion
      - '*'
```
I don't think we need two `*` here. I'll update this.

.github/workflows/build-and-test.yml
BenjaminRodenberg: I reconsidered this: I'm not sure why I restricted running the tests to `master` and `develop` in the first place. Let's keep it as it is here. I think this is useful :+1:

README.md
```diff
@@ -61,12 +61,13 @@ After cloning this repository and switching to the root directory (`fenics-adapt
...
 
-You can run the other tests via `python3 setup.py test`.
+You can run the other tests via `tox`.
+If you want to run them separately, use `tox -e unit` and `tox -e integration` respectively.
 
 Single tests can be also be run. For example the test `test_vector_write` in the file `test_write_read.py` can be run as follows:
 
 ```bash
-python3 -m unittest tests.test_write_read.TestWriteandReadData.test_vector_write
+python3 tox -e unit -- tests/test_write_read.py::TestWriteandReadData::test_vector_write
```
BenjaminRodenberg:
```suggestion
tox -e unit -- tests/test_write_read.py::TestWriteandReadData::test_vector_write
```
I tried this command locally. I think you can skip the `python3`.

BenjaminRodenberg (changes requested):
I think there is a minor thing left. But besides that everything works very well and is a super useful (and extremely overdue) update for this repository. Thank you very much :)

Where I'm still struggling:

If I run the following command:
```
tox -e unit -- tests/test_write_read.py::TestWriteandReadData::test_vector_write
```
I get the following output:
```
unit: commands[0]> pytest tests/unit
==================================================================================================================== test session starts =====================================================================================================================
platform linux -- Python 3.12.3, pytest-8.3.5, pluggy-1.5.0
cachedir: .tox/unit/.pytest_cache
rootdir: /home/benjamin/Programming/fenics-adapter
configfile: pyproject.toml
collected 5 items                                                                                                                                                                                                                                            

tests/unit/test_adapter_core.py ..                                                                                                                                                                                                                     [ 40%]
tests/unit/test_checkpointing.py ...                                                                                                                                                                                                                   [100%]

====================================================================================================================== warnings summary ======================================================================================================================
../../../../usr/lib/python3/dist-packages/dijitso/__init__.py:19
  /usr/lib/python3/dist-packages/dijitso/__init__.py:19: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
    from pkg_resources import get_distribution

../../../../usr/lib/python3/dist-packages/pkg_resources/__init__.py:2871
  /usr/lib/python3/dist-packages/pkg_resources/__init__.py:2871: DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('mpl_toolkits')`.
  Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
    declare_namespace(pkg)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=============================================================================================================== 5 passed, 2 warnings in 5.09s ================================================================================================================
  unit: OK (5.75=setup[0.07]+cmd[5.68] seconds)
  congratulations :) (5.87 seconds)
```

This looks to my like all tests in `unit` are run. However, actually I would like to only see a single test being run. https://stackoverflow.com/a/49935307 gives already a hint: You need `pytest {posargs}` somewhere in the commands. I personally would not start duplicating this in `unit` and `integration`. Instead, it would probably make sense to have a third environment (`all`?) where we run all tests by default and apply the filtering only if given by the user. The command that we run from the command line would look like:
```
tox -e all -- tests/test_write_read.py::TestWriteandReadData::test_vector_write
```

README.md
BenjaminRodenberg: Alternatively, we can run the following:
```
pytest -- tests/integration/test_write_read.py::TestWriteandReadData::test_vector_write
```
As far as I understand we will not automatically use the virtual environment created by tox in this case. But we can also assume that this is a command mainly used by developers and they have an appropriate virtual environment in place. Note that with `tox devenv .venv` you can create the environment quite conveniently. You can also provide a comment in the `README.md` on this fact.

The problem with trying to put this into tox: If we want to have an additional environment `all` that does not differentiate between `unit` and `integration` we will run all tests twice when running `tox` without any arguments: Once for `all` and once for `unit` and `integration`. I think this is not worth the pain.

tox.ini
```diff
@@ -0,0 +1,29 @@
...
+skip_install = True
+description = base environment for testing
+deps =
+    pytest>=8
+    numpy>=1.13.3,<2
+    scipy<1.15.0
+    mpi4py<4
+commands =
+    pipinstall --no-deps
+    pytest {posargs:tests}
```
BenjaminRodenberg, line 16:
```suggestion
    pytest {posargs:tests}
```
Is this line needed?

BenjaminRodenberg:
```suggestion
commands =
    pipinstall --no-deps
    pytest {posargs:tests}
```
Are these lines actually needed?

giuliatum: Fixed in 6628bb81b05a107c68b2e9c0b5893fefba81bc9a

README.md
giuliatum: Fixed in https://github.com/precice/fenics-adapter/commit/6628bb81b05a107c68b2e9c0b5893fefba81bc9a

BenjaminRodenberg (approved):
Looks great! Thank you very much and congratulations to your first contribution :tada:

PR #199, opened 2025-03-31
Fix workflow_dispatch for tag

BenjaminRodenberg:
* Apply tag for both jobs
* Set default if not run via `workflow_dispatch`

### DIFF
--- a/.github/workflows/run-tutorials.yml
+++ b/.github/workflows/run-tutorials.yml
@@ -55,7 +55,7 @@ jobs:
   run_ht_complex:
     name: Run HT, complex
     runs-on: ubuntu-latest
-    container: precice/precice:nightly
+    container: precice/precice:${{ inputs.docker_image_tag }}
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v2
--- a/.github/workflows/run-tutorials.yml
+++ b/.github/workflows/run-tutorials.yml
@@ -19,7 +19,7 @@ jobs:
   run_ht_simple:
     name: Run HT, simple
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


IshaanDesai (commented):
The default tag is `nightly`, so why do we need to additionally mention this now? Also, having such a tag specification is only relevant for release testing, right? Or is there some other motivation?

BenjaminRodenberg:
> The default tag is `nightly`, so why do we need to additionally mention this now?

What do you mean? Be aware that if a workflow is triggered by a push that this does not pick up the defaults from the workflow dispatch. Meaning: [this line](https://github.com/precice/fenics-adapter/blob/725e2a45e00fe3a091ff7803f7ef65f6dfd898da/.github/workflows/run-tutorials.yml#L22) evaluates to `precice/precice:` because `${{ inputs.docker_image_tag }}` is undefined.

> Also, having such a tag specification is only relevant for release testing, right? Or is there some other motivation?

Yes. That's the only reason.

IshaanDesai:
> What do you mean? Be aware that if a workflow is triggered by a push that this does not pick up the defaults from the workflow dispatch.

Ah okay. I was missing this information about the function of how the default works :+1:

IshaanDesai (approved):

BenjaminRodenberg:
> > What do you mean? Be aware that if a workflow is triggered by a push that this does not pick up the defaults from the workflow dispatch.
> 
> Ah okay. I was missing this information about the function of how the default works 👍

It's also 100% not intuitive. If you want to convince yourself that this is not your fault: Take a look at https://github.com/orgs/community/discussions/26322 :sweat_smile:

PR #210, opened 2026-08-03
Make check-links.yml workflow manual

MakisH:
The `check-links.yml` workflow is quite fragile, as pages accessible to humans can (temporarily) be unreachable or block connections from GitHub. While we previously had such a workflow in the website repository, this was causing similar issues there.

As a workaround, I suggest keeping the workflow, but only allowing manual execution. This could be useful to discover some broken links, but would not interfere with the rest of the CI.


IshaanDesai (approved):
Makes sense.

